PR #3, opened 2025-08-02
Allow disabling of default keybind. Fix Readme

AmaliaHaible:
I didnt want to change too much about the readme at first, but since i would have had to add my changes in 3 places, i thought this would be better


dmtrKovalenko:
Thanks but I think these default keymaps should be off these is some very early AI generated BS, thanks for the fix I'll update the PR

dmtrKovalenko:
Sorry I screwed up the PR becuase of dumb github ruling wiht the force pushes but I anyway wanted to rework it so I counted you as a contributor in the other PR and closing this one 

https://github.com/dmtrKovalenko/fff.nvim/pull/26

PR #4, opened 2025-08-02
Fixed prompt not getting wiped on esc

AmaliaHaible:
When pressing escape with something in the prompt, it counted as a modified buffer, so i couldnt quit with :q, only with :qa!. This fixes it, but im not sure if this is the best way to do it


Tebro:
This happens to me even if I don't use esc. So if I use fff at all I end up with the prompt buffer where I cant directly access the commandline to do :q! but instead have to use fff to navigate to some other file and then do :qa! which isn't ideal. I'll test this branch and report back if it fixes this for me as well.

Tebro:
Testing report: this PR fixes the issue for me.

AmaliaHaible:
This is also fixed by #10 which cleans up the buffers and exits insert mode as well. You should be able to go back to the main branch.

PR #5, opened 2025-08-02
feat: support nixos

fvrings:
add support for nixos
one thing I'm not certain is the usage of build.rs
I can not find references of `target/release/version`
Is this file for future development?


saghen:
That file is used by blink.cmp in its build process so it knows which version the library was built for. It's safe to remove here.

PR #7, opened 2025-08-02
Added main config to picker UI

dwoznicki:
The picker UI wasn't reading the options passed in at setup time, making it difficult to adjust picker width, height, etc. Better to include them, but allow the `open(opts)` options to overwrite them.


dmtrKovalenko:
Appreciate your fix but is already obsolete by another PR went in. Anyway thank you so much

PR #12, opened 2025-08-02
style: Removing Tokio as a dependency & addressing Clippy lints

dandxy89:
Saw this announced on Reddit - thank you!

### DIFF
--- a/lua/fff/rust/file_picker.rs
+++ b/lua/fff/rust/file_picker.rs
@@ -180,7 +180,7 @@ impl std::fmt::Debug for FilePicker {
         f.debug_struct("FilePicker")
             .field("base_path", &self.base_path)
             .field("git_workdir", &self.git_workdir)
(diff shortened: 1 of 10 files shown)


lua/fff/rust/file_picker.rs
```diff
@@ -317,11 +318,12 @@ impl FilePicker {
 
                 file.update_frecency_scores();
             }
-        };
+        }
 
-        Ok(self.get_cached_files())
+        self.get_cached_files()
```
dmtrKovalenko, line 323: just for reference sometimes I leave the functions return result if I know that I might need result later, it is anyway fused by a compiler if there a single pass used

dmtrKovalenko:
I wasn't able to push into your branch so I made another PR with your commits preserved https://github.com/dmtrKovalenko/fff.nvim/pull/24

PR #25, opened 2025-08-02
feat: add configurable sibling file prioritization with performance optimizations

sQVe:
## Summary

Enhance the fuzzy file picker with sibling file prioritization that considers your current context. When editing a file, the picker now prioritizes related files (similar names, nearby directories) while maintaining performance through optimized scoring algorithms.

This helps with common workflows where you need to quickly navigate between related files (e.g., `user.ts` and `user.test.ts`, or files in the same feature directory) by making the most relevant results appear first.

## Key Changes

- **Jaro-Winkler Similarity Scoring**: Add string similarity algorithm to identify related filenames with configurable thresholds
- **Performance-Optimized Scoring**: Gate expensive similarity calculations behind distance checks to maintain fast response times
- **Configurable Preference System**: New `scoring.same_dir_preference` option (0.0-1.0) to control prioritization strength
- **Pre-computed Path Data**: Cache frequently accessed path components to reduce redundant calculations
- **Comprehensive Documentation**: Updated README and Vim help with usage examples and configuration guidance

## Configuration

Users can now control file prioritization behavior:

```lua
require("fff").setup({
  scoring = {
    same_dir_preference = 0.7,  -- Range: 0.0 (no preference) to 1.0 (strong preference)
  },
})
```

## Performance

- Similarity calculations are only performed for files within `MAX_DISTANCE_FOR_SIMILARITY_BONUS` directory levels (default: 2)
- Path components are pre-computed and cached in `CurrentFileData` struct
- Jaro-Winkler calculations use 0.5 similarity threshold to balance relevance and performance

## Test Plan

- [ ] Verify configuration validation with invalid values (< 0.0 or > 1.0)
- [ ] Test file prioritization with different `same_dir_preference` values
- [ ] Confirm related files (similar names) appear higher in results
- [ ] Validate performance with large codebases (no noticeable lag)
- [ ] Check documentation examples work as described


dmtrKovalenko:
wow this is huge and makes a lot of sense, I'll give a proper review for this feature tomorrow! Thanks for a huge addition

sQVe (commented):
-

lua/fff/rust/file_picker.rs
```diff
@@ -244,12 +247,26 @@ impl FilePicker {
         let total_files = sync_data.files.len();
 
         // small queries with a large number of results can match absolutely everything
-        let max_typos = (query.len() as u16 / 4).clamp(2, 6);
+        let max_typos = if query.len() <= 4 {
+            0
+        } else {
+            (query.len() as u16 / 5).clamp(1, 3)
+        };
```
sQVe, line 254: I found the previous `max_typos` value to be too permissive, which was problematic when scoring against proximity. I primarily work within a large TypeScript monorepo, and when typing `Page` it didn't filter out any items.

This change should hopefully be a bit more adaptive. I wouldn't be surprised if we would need to tweak this though.

lua/fff/rust/path_utils.rs
```diff
@@ -1,90 +1,333 @@
...
+const MAX_PENALTY_LEVEL_MULTIPLIER: i32 = 10;
+
+pub fn calculate_filename_similarity_bonus(
+    current_file_path: &str,
+    candidate_file_path: &str,
+    max_bonus: i32,
+    similarity_threshold: f64,
+) -> i32 {
+    use std::path::Path;
+    use strsim::jaro_winkler;
```
sQVe, line 10: I think Jaro Winkler is a good choice since:
1. Prefix bias: Gives higher scores to strings that share common prefixes, which aligns with how developers typically name related files (e.g., `user.rs`, `user_test.rs`, `user_service.rs`)
  2. Typo tolerance: Handles minor typos and variations gracefully, which is important for fuzzy file finding
  3. Performance: It's computationally efficient compared to algorithms like Levenshtein.
  4. Score range: Returns normalized scores between 0.0 and 1.0, making it easy to set meaningful thresholds

lua/fff/config_utils.lua
```diff
@@ -0,0 +1,44 @@
...
+end
+
+--- @param preference number User preference value between 0.0 and 1.0
+--- @return table Internal scoring parameters
+function M.map_preference_to_scoring(preference)
+  return {
+    directory_distance_penalty = -8, -- Balanced penalty for different directories
+    filename_similarity_bonus_max = math.floor(50 * preference), -- Moderate sibling bonus (35 with default 0.7)
+    filename_similarity_threshold = 0.5, -- Good relevance/performance balance
+    max_search_directory_levels = math.floor(1 + 3 * preference),
```
sQVe, line 40: I got these defaults by forcing multiple AI models to reach a consensus on what they believed was the optimal values. I then manually tested and iterated multiple times, feeding the results to the models.

With that said, we should really see these values as an initial best guess. My gut feeling is that we will lower these values rather than raising them in the future.

sQVe (commented):
👋🏼 

I think this project has great potential, and it feels very snappy 👏🏼. I'm currently using **snacks.nvim** with a [custom multi picker](https://github.com/sQVe/dotfiles/blob/6730231aeb838c36f43659818bd9de1c52e1c606/config/nvim/lua/sQVe/plugins/snacks/picker.lua#L55-L86), which provides basic proximity handling for the active buffer.

This is really essential when working within large monorepos, and projects that is very structure / name oriented - like TypeScript & Next.js etc.

---

Do note that I'm very novice when it comes to Rust, so be thorough and kind. I'm more than happy to make whatever changes you feel are necessary to get this in! ❤️

lua/fff/rust/file_picker.rs
dmtrKovalenko: I think we should probably make it configurable as well while allowing typos starting from the very start. I make typos even in 4 lines and I find it absolutely useful to have them

lua/fff/rust/lib.rs
```diff
@@ -76,7 +92,18 @@ pub fn fuzzy_search_files(
         .as_ref()
         .ok_or_else(|| Error::InvalidPath("File picker not initialized".to_string()))?;
 
-    let results = picker.fuzzy_search(&query, max_results, max_threads, current_file.as_ref());
+    let distance_penalty = distance_penalty.unwrap_or(-8);
+    let relation_bonus_max = relation_bonus_max.unwrap_or(35);
```
dmtrKovalenko, line 96: I don't think we should give a default dardcoded bonusws or penalties

dmtrKovalenko (commented):
Sorry for not getting your review fast enough. I'll spend a few days working with this version to see how it feels and will get back to you. Thank you very much this is an amazing contribution!

lua/fff/rust/lib.rs
sQVe: Do you think this should be configurable instead, or solved in some other way?

sQVe:
@dmtrKovalenko No need to rush ❤️ I understand if you want to make sure this feels right and that you might want to get the basics right in the picker - before adding further tweaks.

I prefer having a picker that solves proximity, open buffers, and fuzzy finding - but that might not be the goal of this project.

dmtrKovalenko:
It actually is I am really grateful for your input I just want to double check that it doesn't break the existing flows and maybe tweak/make configurabel the scoring impact. We are going to merge this PR anyway

PR #35, opened 2025-08-03
feat(image): add terminal support check via snacks

rachartier:
Hello,

Ghostty and Wezterm also supports displaying images with Snacks.image and with Kitty's kitten icat. 

I think it is better to simply call `snacks.image` without even checking if the terminal is supported, as Snacks already does this: https://github.com/folke/snacks.nvim/blob/main/lua/snacks/image/buf.lua#L9 with https://github.com/folke/snacks.nvim/blob/main/lua/snacks/image/init.lua#L212.

So it is quite redundant to do this check again, as you verify exactly this with `success` and `placement` variable in your `M.display_image` function.

What do you think?

### DIFF
--- a/lua/fff/file_picker/image.lua
+++ b/lua/fff/file_picker/image.lua
@@ -41,7 +41,6 @@ local IMAGE_EXTENSIONS = {
   '.avif',
 }
 
---- Check if file is an image
 --- @param file_path string Path to the file
 --- @return boolean True if file is an image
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
I wasn't able to push into your repository so I created another PR with your commit preserved https://github.com/dmtrKovalenko/fff.nvim/pull/50

PR #40, opened 2025-08-03
fix: Support macOS .dylib extension in release script

vadymbiliuk:
## Motivation

On macOS, Rust `cdylib` outputs are named `*.dylib`, not `*.so`. So it's impossible to build with nix-darwin machines. This change fixes it.


dmtrKovalenko:
thanks!

PR #41, opened 2025-08-03
security: fix shell injection vulnerabilities in file path handling

sQVe:
## Summary

Fixes shell injection vulnerabilities in file path handling where malicious file names could potentially execute arbitrary commands.

## What was the issue?

A few system command calls were using unescaped file paths, which could be problematic if someone created files with special shell characters in their names.

For example, a file named `image.jpg; rm -rf /` could cause issues when the picker tries to preview it.

## What changed?

Updated **5 system command calls** across 2 files to use `vim.fn.shellescape()` for proper path escaping:

##### `lua/fff/file_picker/image.lua`
- Fixed `file` and `identify` commands in `get_image_dimensions()`

##### `lua/fff/file_picker/preview.lua`  
- Fixed `tail` command in `read_file_tail()`
- Fixed `file` and `xxd` commands in `preview_binary_file()`

**The fix**:
```lua
-- Before
string.format('command "%s"', file_path)

-- After  
string.format('command %s', vim.fn.shellescape(file_path))
```

## Testing

- Verified the picker still works normally with regular files
- Tested with files that have special characters in their names
- Image previews and binary file handling work as expected

### DIFF
--- a/lua/fff/file_picker/image.lua
+++ b/lua/fff/file_picker/image.lua
@@ -158,7 +158,7 @@ end
 --- @return number|nil, number|nil Width and height in pixels
 function M.get_image_dimensions(file_path)
   -- Try file command first
-  local cmd = string.format('file "%s"', file_path)
+  local cmd = string.format('file %s', vim.fn.shellescape(file_path))
   local result = vim.fn.system(cmd)
 
   if vim.v.shell_error == 0 then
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
thank you for noting and fixing this, this is amazing!

PR #42, opened 2025-08-04
fix: use correct path separators to check for .git folder on windows

### LINKED ISSUE #38: Bug: .git Folder gets indexed on Windows
Using the default settings on Windows, i can also see all the files in my .git folder. This does not happen with the same config in the same folder via WSL

Windows:
Image

WSL:
Image

WillLillis:
Thanks so much for this awesome tool, I've really enjoyed using it the last few days!

The first two commits are just some minor style things I noticed while reading through the project, happy to revert if they're not desired.

The last commit fixes the problem with #38, the issue was just the difference in path separators. I opted to use a conditional `cfg!` rather than `std::path::MAIN_SEPARATOR` in a `format!` string for performance reasons. I've tested this patch out locally on a windows and linux machine.

- Closes #38

### DIFF
--- a/lua/fff/rust/file_picker.rs
+++ b/lua/fff/rust/file_picker.rs
@@ -687,7 +687,13 @@ fn update_git_status_for_paths(
 
 #[inline]
 fn is_git_file(path: &Path) -> bool {
-    path.to_str().is_some_and(|path| path.contains("/.git/"))
+    path.to_str().is_some_and(|path| {
+        if cfg!(target_family = "windows") {
+            path.contains("\\.git\\")
+        } else {
+            path.contains("/.git/")
+        }
+    })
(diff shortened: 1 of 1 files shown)


lua/fff/rust/file_picker.rs
```diff
@@ -689,7 +687,13 @@ fn update_git_status_for_paths(
 
 #[inline]
 fn is_git_file(path: &Path) -> bool {
-    path.to_str().is_some_and(|path| path.contains("/.git/"))
+    path.to_str().is_some_and(|path| {
+        if cfg!(target_family = "windows") {
```
dmtrKovalenko, line 691: didn't even know this is an option lol have always used target_os

PR #44, opened 2025-08-04
feat: manage openssl with cargo

### LINKED ISSUE #43: openssl-sys requires linking against system openssl library
I have installed openssl lib using nix on macOS, however openssl-sys package has a trouble finding the library. I suggest we add openssl as a cargo dependency and enable the "vendored" feature to make this build more self-contained. This also helps users that don't have openssl headers installed.

madmaxieee:
List openssl as a cargo dependency to avoid relying on system library. This makes the build not dependent of system openssl library, making it more self-contained.

fix #43

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,6 +238,7 @@ dependencies = [
  "neo_frizbee",
  "notify",
  "notify-debouncer-full",
+ "openssl",
  "pathdiff",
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
oh my god I didn't even notice that libgit2 depends on ssh thanks for the fix

PR #51, opened 2025-08-04
Fix watcher leak in picker

### LINKED ISSUE #49: Potential memory leak
I had nvim on in the background and my computer got quite sluggish, nvim was using over 20GB of memory. Is there any way to debug if that leak comes from fff?

Tebro:
Closes #49

### DIFF
--- a/lua/fff/picker_ui.lua
+++ b/lua/fff/picker_ui.lua
@@ -989,6 +989,8 @@ function M.close()
     if buf and vim.api.nvim_buf_is_valid(buf) then vim.api.nvim_buf_delete(buf, { force = true }) end
   end
 
+  main.stop_background_monitor()
+
   M.state.input_win = nil
   M.state.list_win = nil
   M.state.file_info_win = nil
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -449,4 +449,8 @@ function M.change_indexing_directory(new_path)
   return true
 end
 
+function M.stop_background_monitor()
+  pcall(fuzzy.stop_background_monitor)
+end
+
 return M


dmtrKovalenko (approved):

lua/fff/picker_ui.lua
```diff
@@ -989,6 +989,8 @@ function M.close()
     if buf and vim.api.nvim_buf_is_valid(buf) then vim.api.nvim_buf_delete(buf, { force = true }) end
   end
 
+  main.stop_background_monitor()
```
dmtrKovalenko, line 992: This is not an option to do because we have to keep the sync running in the background, we should clean it up when the neovim is requested to exit but I've been sure that it should already be cleaned up let me research this a little bit

Tebro: Yeah the problem is that it seems to spawn multiple of these processes (one on each use of the finder). And they never die until the editor is closed. 

I do however now wonder if this actual is a valid fix, cause now I have an instance taking 4GB of memory again. 

This is a bit hard to verify if a fix is working or not.

dmtrKovalenko: they spawning once per neovim session (at least they should be) I will try to wrap my head around this more deeply can you share the platform info

Tebro: I am on Linux, kernel 6.15.8, alacritty terminal, Neovim v0.11.3

dmtrKovalenko:
I debugged it a little bit more and apperently I misread the docs of notify rs and missed the idea that they spawn their own debouncer thread which led to the hanging threads by debouncer. After this PR everything should be good https://github.com/dmtrKovalenko/fff.nvim/pull/56

Issue #32, opened 2025-08-03
Error: improve UI when preview.enable = false


jonahfang:
see pic below, even I disable preview, the border does not hide:

<img width="1935" height="543" alt="Image" src="https://github.com/user-attachments/assets/ca4d0f8d-9589-48f9-8410-f38ac7fba929" />

dinhmai74:
fixed in [#18](https://github.com/dmtrKovalenko/fff.nvim/pull/18) it should worked now with this setup  @jonahfang 

```
require("fff").setup({
  preview = {
      enabled = false,
  },
})
```

jonahfang:
But I am using the latest version: `commit d3e34a78022c5ea3bc7d14eb7ac2b801c57abafd (HEAD -> main, origin/main, origin/HEAD)`

dmtrKovalenko:
Might it be related to enabled debug scoring? Try <F2> but I will anyway try to look into it

dinhmai74:
<img width="1472" height="884" alt="Image" src="https://github.com/user-attachments/assets/847dd5f8-9c53-46e7-8357-fdaacf6868bc" />

can you help share the config @jonahfang , I got this with this minimal config 

```
{
  'dmtrKovalenko/fff.nvim',
  build = 'cargo build --release',
  preview = {
    enabled = false,
  },
}

```

jonahfang:
I have found the error will corrected if I changed the code in lua/fff/picker_ui.lua like this:

````
  if M.enabled_preview() then
      if debug_enabled then
        M.state.file_info_win = vim.api.nvim_open_win(M.state.file_info_buf, false, {
          relative = 'editor',
          --width = preview_width,
          width = list_width,
          height = file_info_height,
          col = col + list_width + 3,
          row = row + 1,
          border = 'single',
          style = 'minimal',
          title = ' File Info ',
          title_pos = 'left',
        })
      else
        M.state.file_info_win = nil
      end
  else
    M.state.file_info_win = nil
  end
```

Issue #29, opened 2025-08-02
Error when show_scores = true!


jonahfang:
debug = {
                show_scores = true,
            },

E5108: Error executing lua: ...nvim/data/confignext/lazy/fff.nvim/lua/fff/picker_ui.lua:91: 'width' key must be a positive Integer

PR #55, opened 2025-08-04
Fix/issue#32 should not show debug preview pannel when preview is disable

dinhmai74:
- Fix: state.config should merge main.config and manual options.
- Fix: Do not create the debug panel when the enable_preview option is falsy.

https://github.com/user-attachments/assets/1a513c94-9e2c-49ce-b2c5-8b786e5c87a1

Resolve: #32


dinhmai74:
Also resolve #29

dmtrKovalenko:
The only quirk of this PR is that when you have preview disabled you don't have the debug window at all but I think I can adress this later I suppose mostl people will use preview

PR #60, opened 2025-08-04
Fix: vim.notify() Expects string

htmluz:
On NVIM v0.11.3 and Lua 5.1 errors were thrown when you tried to scroll in a preview.
It was caused by `vim.notify(win_height)` and `tostring()` solves the problem. 

image


lua/fff/file_picker/preview.lua
```diff
@@ -478,7 +478,7 @@ function M.scroll(lines)
   if not M.state.winid or not vim.api.nvim_win_is_valid(M.state.winid) then return end
 
   local win_height = vim.api.nvim_win_get_height(M.state.winid)
-  vim.notify(win_height)
+  vim.notify(tostring(win_height))
```
dmtrKovalenko:
```suggestion
```

sorry this a debug line we should remove it

PR #62, opened 2025-08-04
fix: eliminate visual flashes and unnecessary rerenders in file picker

sQVe:
### Summary

Fixes the visual flicker when opening the file picker and improves performance by eliminating unnecessary rerenders and polling overhead.

### What Changed

**🎯 Atomic Rendering & Performance**
- Reduced picker opening from 3+ renders to 1 single render
- Eliminated the empty → files → files+git visual progression
- Removed continuous 500ms polling loop that was consuming CPU cycles
- Added 500ms timeout so it won't hang on slow filesystems

**⚡ Caching Improvements**
- Cache icon provider failures to prevent repeated setup attempts and spam warnings
- Eliminates redundant provider lookups on every picker open
- Restore frecency cache sync that got missed during refactor

### Performance Impact

- **CPU Usage**: Removed background polling/monitoring that ran every 500ms
- **Render Cycles**: Down from 3+ renders to 1 atomic render per picker open
- **Setup Overhead**: Icon provider failures now cached instead of retried repeatedly

### How to Test

- [ ] Try the old version - notice flickering and multiple UI updates
- [ ] Try this branch - should pop up smoothly in one render
- [ ] Test on a slow/large repo - should show a warning after 500ms but still work
- [ ] Verify file rankings update properly after accessing files
- [ ] Check that failed icon providers don't spam warnings anymore

### Technical Notes

Uses atomic rendering (wait-then-show) instead of progressive loading, following the pattern used by VSCode and other modern editors. Primarily removed problematic polling code rather than adding complexity.

### DIFF
--- a/lua/fff/file_picker/init.lua
+++ b/lua/fff/file_picker/init.lua
@@ -187,15 +187,16 @@ function M.is_initialized() return M.state.initialized end
 function M.get_config() return M.config end
 
 --- Get scan progress information
---- @return table Progress information with total_files, scanned_files, is_scanning
+--- @return table Progress information with scanned_files_count, is_scanning
 function M.get_scan_progress()
-  if not M.state.initialized then return { total_files = 0, scanned_files = 0, is_scanning = false } end
+  if not M.state.initialized then return { total_files = 0, scanned_files_count = 0, is_scanning = false } end
 
   local ok, result = pcall(fuzzy.get_scan_progress)
   if not ok then
     vim.notify('Failed to get scan progress: ' .. result, vim.log.levels.WARN)
-    return { total_files = 0, scanned_files = 0, is_scanning = false }
+    return { scanned_files_count = 0, is_scanning = false }
   end
+
   return result
 end
 
--- a/lua/fff/file_picker/init.lua
+++ b/lua/fff/file_picker/init.lua
@@ -154,7 +154,11 @@ function M.access_file(file_path)
   if not M.state.initialized then return end
 
   local ok, result = pcall(fuzzy.access_file, file_path)
-  if not ok then vim.notify('Failed to record file access: ' .. result, vim.log.levels.WARN) end
+  if not ok then
+    vim.notify('Failed to record file access: ' .. result, vim.log.levels.WARN)
+  else
+    pcall(fuzzy.refresh_git_status)
+  end
 end
 
 --- Get file content for preview
--- a/lua/fff/file_picker/init.lua
+++ b/lua/fff/file_picker/init.lua
@@ -154,11 +154,7 @@ function M.access_file(file_path)
   if not M.state.initialized then return end
 
   local ok, result = pcall(fuzzy.access_file, file_path)
-  if not ok then
-    vim.notify('Failed to record file access: ' .. result, vim.log.levels.WARN)
-  else
-    pcall(fuzzy.refresh_git_status)
-  end
+  if not ok then vim.notify('Failed to record file access: ' .. result, vim.log.levels.WARN) end
 end
 
 --- Get file content for preview
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -141,6 +141,7 @@ function M.setup_global_autocmds()
             if stat and stat.type == 'file' then
               local relative_path = vim.fn.fnamemodify(file_path, ':.')
               pcall(fuzzy.access_file, relative_path)
+              pcall(fuzzy.refresh_git_status)
             end
           end)
         end
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -141,7 +141,7 @@ function M.setup_global_autocmds()
             if stat and stat.type == 'file' then
               local relative_path = vim.fn.fnamemodify(file_path, ':.')
               pcall(fuzzy.access_file, relative_path)
-              pcall(fuzzy.refresh_git_status)
+              pcall(fuzzy.update_single_file_frecency, relative_path)
             end
           end)
         end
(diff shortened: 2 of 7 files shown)


lua/fff/main.lua
```diff
@@ -141,6 +141,7 @@ function M.setup_global_autocmds()
             if stat and stat.type == 'file' then
               local relative_path = vim.fn.fnamemodify(file_path, ':.')
               pcall(fuzzy.access_file, relative_path)
+              pcall(fuzzy.refresh_git_status)
```
dmtrKovalenko: Refresh git status is a costly operation counting all the files, why do we need to call it on every file open?

lua/fff/rust/lib.rs
```diff
@@ -183,14 +182,7 @@ pub fn wait_for_initial_scan(_: &Lua, timeout_ms: Option<u64>) -> LuaResult<bool
...
 
-    let timeout = Duration::from_millis(timeout_ms.unwrap_or(5000)); // Default 5s timeout
-    let start_time = std::time::Instant::now();
-
-    while picker.is_scan_active() && start_time.elapsed() < timeout {
-        std::thread::sleep(Duration::from_millis(50));
-    }
-
-    Ok(!picker.is_scan_active())
+    Ok(picker.wait_for_complete_scan(timeout_ms))
```
dmtrKovalenko: This makes sense but I think we should just inline this function

dmtrKovalenko (commented):
Thank you for your collaborations, this is amazing! this is definitely a problem that status flashes after you open a picker but this should be solved on the file watcher side. 

We should not traverse the whole dir when we have the notifier on every file action.

lua/fff/main.lua
sQVe: 🙈 

This is the result of a lazy man being tired before going to bed 😆. I added this here as a dumb patch for the frecency not being updated correctly. 

I will try to fix this correctly instead! 👍🏼

sQVe: I think I have resolved this with https://github.com/dmtrKovalenko/fff.nvim/pull/62/commits/ce52a639a3fa0f936bde6de5e3402967cf8ffc12. Please take a look! 👍🏼

lua/fff/rust/lib.rs
sQVe: By inline, are you referring to adding a `#[inline]` attribute or extracting the logic from `wait_for_complete_scan` to `wait_for_initial_scan`?

I'm assuming you're referring to the latter.

lua/fff/main.lua
```diff
@@ -141,6 +141,7 @@ function M.setup_global_autocmds()
             if stat and stat.type == 'file' then
               local relative_path = vim.fn.fnamemodify(file_path, ':.')
               pcall(fuzzy.access_file, relative_path)
+              pcall(fuzzy.update_single_file_frecency, relative_path)
```
dmtrKovalenko: this makes much more sense https://github.com/dmtrKovalenko/fff.nvim/blob/main/lua/fff/rust/lib.rs#L118-L118 but instead of 2 different calls I think we should just add it here https://github.com/dmtrKovalenko/fff.nvim/blob/main/lua/fff/rust/lib.rs#L118-L118 besides that everything seems to be good to go

lua/fff/picker_ui.lua
```diff
@@ -948,6 +948,7 @@ function M.select(action)
 
   local relative_path = vim.fn.fnamemodify(item.path, ':.')
   file_picker.access_file(relative_path)
+  file_picker.refresh_git_status()
```
dmtrKovalenko: oh and here ^ I don't think we need to refresh the git status on selection either, when user selects a file it doesn't mean that the file is going to change git status

lua/fff/main.lua
sQVe: This should be resolved with https://github.com/dmtrKovalenko/fff.nvim/pull/62/commits/6d9ef6adf8c7f0e67b7aa493297dd99eaba166ac.

lua/fff/picker_ui.lua
sQVe: This should be resolved with https://github.com/dmtrKovalenko/fff.nvim/pull/62/commits/6d9ef6adf8c7f0e67b7aa493297dd99eaba166ac.

lua/fff/file_picker/init.lua
```diff
@@ -154,7 +154,11 @@ function M.access_file(file_path)
   if not M.state.initialized then return end
 
   local ok, result = pcall(fuzzy.access_file, file_path)
-  if not ok then vim.notify('Failed to record file access: ' .. result, vim.log.levels.WARN) end
+  if not ok then
+    vim.notify('Failed to record file access: ' .. result, vim.log.levels.WARN)
+  else
+    pcall(fuzzy.refresh_git_status)
```
dmtrKovalenko: this should also not happen on every access

lua/fff/rust/lib.rs
```diff
@@ -224,6 +255,10 @@ fn create_exports(lua: &Lua) -> LuaResult<LuaTable> {
         "refresh_git_status",
         lua.create_function(refresh_git_status)?,
     )?;
+    exports.set(
```
dmtrKovalenko: I don't think we need to expose it to the lua API, this is more of an internal API

lua/fff/file_picker/init.lua
sQVe: Grrr... I'll be more thorough next time 😊

dmtrKovalenko:
Okay so I a little bit restructured the code to eliminate the blocking on the UI. This 500ms of neovim is actually noticable, we should avoid any visible blocking. Instead I restored back the monitor for status event but made it more snappier so it is waiting on neovim side to wait for 100ms and then increasing the waiting time if we hit a very large directory

With your current changes very large directory just stuck in the never updated state


https://github.com/user-attachments/assets/131e376d-de7f-47c0-bb77-92b233efcaaf

PR #73, opened 2025-08-06
show relative path in `:buffers`

sibouras:
when pressing `:b` then tab show the relative path instead of full path

### DIFF
--- a/lua/fff/picker_ui.lua
+++ b/lua/fff/picker_ui.lua
@@ -935,15 +935,14 @@ function M.select(action)
   vim.cmd('stopinsert')
   M.close()
 
-  local file_path = item.path
   if action == 'edit' then
-    vim.cmd('edit ' .. vim.fn.fnameescape(file_path))
+    vim.cmd('edit ' .. vim.fn.fnameescape(relative_path))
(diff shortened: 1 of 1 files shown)


sQVe:
Can you explain what the use case is here? What does it fix?

dmtrKovalenko:
I think the change makes sense currently we are openning full path when the changing the file in the active buffer but I am not sure this is a good idea to fix because base path of the picker could be different from your vim cwd

sibouras:
before this PR when you open files with the default `:e`, mini.files, telescope... they all show up as relative paths in `:ls` unlike fff.nvim


https://github.com/user-attachments/assets/cfdb023e-ccfc-4ba7-93c1-5cd3dc2b754a

also on windows, when you open buffers as full paths and you exit nvim it crashes the terminal, its related to how paths are saved in the shada file(it's a long standing issue on windows) but we can avoid it if we open them as relative.

> but I am not sure this is a good idea to fix because base path of the picker could be different from your vim cwd

`vim.fn.fnamemodify(item.path, ':.')` doesn't change the path if that's the case

dmtrKovalenko (approved):

PR #81, opened 2025-08-09
feat: Rework picker allocations and state management

dmtrKovalenko:
I tried to simplify picker managing the state to make it more composable and debuggable, also this allows to significantly reduce amount of allocations needed and be more smart about handling the fs events

### DIFF
--- a/lua/fff/rust/file_key.rs
+++ b/lua/fff/rust/file_key.rs
@@ -2,3 +2,21 @@
 pub struct FileKey {
     pub path: String,
 }
+
+impl FileKey {
+    pub fn new(path: String) -> Self {
+        Self { path }
+    }
+
+    pub fn into_path_buf(self) -> std::path::PathBuf {
+        std::path::PathBuf::from(self.path)
+    }
+
+    pub fn as_path(&self) -> &std::path::Path {
+        std::path::Path::new(&self.path)
+    }
+
+    pub fn into_string(self) -> String {
+        self.path
+    }
+}
--- a/src/bin/jemalloc_profile.rs
+++ b/src/bin/jemalloc_profile.rs
@@ -1,0 +1,266 @@
+use fff_nvim::{file_picker::FilePicker, FILE_PICKER};
+use std::env;
+use std::thread;
+use std::time::Duration;
+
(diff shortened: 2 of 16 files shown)


src/bin/jemalloc_profile.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    let mut max_allocated = initial_allocated;
+    let mut max_active = initial_active;
+    let mut max_resident = initial_resident;
+
+    for i in 0..iterations {
+        let query = query_pattern(i);
+
+        let (result_count, _total_matched) = {
+            let file_picker_guard = FILE_PICKER.read().unwrap();
```
sQVe, line 84: How do we want handle lock poisoning? We should probably establish a pattern for it.

In this case it would make sense to use the `parking_lot`, no?

sQVe (dismissed):
I think this looks great, just have one minor suggestion. Nothing blocking! 👍🏼 🚀

lua/fff/rust/file_key.rs
```diff
@@ -2,3 +2,21 @@
 pub struct FileKey {
     pub path: String,
 }
+
+impl FileKey {
+    pub fn new(path: String) -> Self {
+        Self { path }
+    }
+
+    pub fn into_path_buf(self) -> std::path::PathBuf {
```
WillLillis, line 11: More a stylistic nit, but why add `into_path_buf` and `into_string` as special member functions instead of implementing the same functionality via `Into`?

dmtrKovalenko: I don't like conversions traits the much, they make sense only if you have a lot of combinations of converting something into something + it is never clear which exact code will be exectued 

I am following this convention https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

src/bin/jemalloc_profile.rs
dmtrKovalenko: I am not sure if partking_lolt will give us any visible benefit it makes sense when you have a lot of concurrent threads running and accessing locks at the same time, we just have a global state that is shared between 2 threads I think std locks are more than enough for this 

but if you are curious to try you can try to benchmark it with parking lot and if it gives a visible benefit we can for sure merge it

PR #86, opened 2025-08-10
feat: Cap filename bonus

dmtrKovalenko:
Noticed the following issue: when the query is large e.g. 300 the 20% of the filename bonus is 

image


sQVe (approved):
LGTM! 🥇

PR #101, opened 2025-08-13
feat: Add neovim package + nix ci workflow

jla2000:
I have added a nix package that can be included into nix-based neovim configs, together with a ci pipeline that checks that the nix flake will not get broken in future.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,31 @@
+name: Nix CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
great thank you for doing this, will this now work with nixvim right?

jla2000:
Yes that should work 👍 
I have not tested it in nixvim though.
It can e.g. be used like this:
https://github.com/jla2000/nvim-bundle/commit/6b6022d62077d7d8f26752e23a699df5db5af068

PR #102, opened 2025-08-13
refactor: better state management

shawnyu5:
Attempt to standardise state management by refactor state management into its own module, and add type hints for the state table. This makes it easier to add additional fields to the state later on.

Also fixed a few places where the user config is not being respected.

### DIFF
--- a/lua/fff/state.lua
+++ b/lua/fff/state.lua
@@ -107,94 +107,97 @@ local M = {}
 ---@field icons ConfigIcons File icon display options
 ---@field debug ConfigDebug Debug options
 
+---@type Config
+local default_config = {
+  base_path = vim.fn.getcwd(),
+  max_results = 100,
+  max_threads = 4,
+  prompt = '🪿 ',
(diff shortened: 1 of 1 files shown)


lua/fff/state.lua
```diff
@@ -0,0 +1,200 @@
...
+---@field image_preview ConfigImagePreview Image preview options
+---@field icons ConfigIcons File icon display options
+---@field debug ConfigDebug Debug options
+
+-- State
+---@type State
+M = {
+  initialized = false,
+  last_scan_time = 0,
+  config = {
```
dmtrKovalenko: can we have the default config isolted somewhere so it is easy to link this from the readme

shawnyu5: Moved the default config into its own variable in the file.

shawnyu5:
Merged main into this branch. Fixed conflicts.

dmtrKovalenko:
I'm sorry this is too outdated

PR #115, opened 2025-08-18
feat: smart lazy-loading + no forced setup

### LINKED ISSUE #114: feature request: smart lazy-loading
Hey 👋 

I'd love to try out your plugin. I gave it a go, but noticed two things that keep me from using it at this time:

- It forces me to call a `setup` function for the plugin to be usable.
- Relies on lazy.nvim for lazy loading.

See also: 

- https://github.com/nvim-neorocks/nvim-best-practices?tab=readme-ov-file#zap-initialization
- https://github.com/nvim-neorocks/nvim-best-practices?tab=readme-ov-file#sleeping_bed-lazy-loading

If you're open to it and you believe it's feasible, I'd be happy to open a PR 😄

mrcjkb:
Closes #114 

I've tested this by adding the plugin's nix flake output to my nvim config.

Things done:

- Split configuration and initialization
- Refactored configuration to allow you to use both `require('fff').setup(config)` and `vim.g.fff = config`.
  The config is validated the first time it is requested internally via `require('fff.conf').get()`.
- Fixed an `attempt to index nil value` error when opening the picker.
- Initialize the plugin lazily. To make it initialize eagerly, set `lazy_sync = false` in the config.
- If initializing eagerly, defer until after `UIEnter` (this is equivalent to initializing on lazy.nvim's `VeryLazy` event).
- Create commands in the `plugin/` script. The commands `require` fff's Lua modules lazily, so this has less startup overhead than would be needed to create lazy-loading configurations with lazy.nvim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,31 +40,59 @@ It comes with a dedicated rust backend runtime that keep tracks of the file inde
 FFF.nvim requires:
 
 - Neovim 0.10.0+
-- Rust toolchain (requires nightly for building the native backend)
+- [Rustup](https://rustup.rs/) (we require nightly for building the native backend rustup will handle toolchain automatically)
 
-### Package Managers
+### Installation
 
 #### lazy.nvim
 
 ```lua
 {
-  "dmtrKovalenko/fff.nvim",
-  build = "cargo build --release",
+  'dmtrKovalenko/fff.nvim',
+  build = 'cargo build --release',
   -- or if you are using nixos
   -- build = "nix run .#release",
   opts = { -- (optional)
-    -- pass here all the options
+    debug = {
+      enabled = true,     -- we expect your collaboration at least during the beta
+      show_scores = true, -- to help us optimize the scoring system, feel free to share your scores!
+    },
   },
   -- No need to lazy-load with lazy.nvim.
   -- This plugin initializes itself lazily.
   lazy = false,
-  init = function()
-    -- try it if you didn't it is a banger keybinding for a picker
-    vim.keymap.set("n", "ff", function()
-      require("fff").find_files() -- or find_in_git_root() if you only want git files
-    end, { desc = "Open file picker" })
-  end,
+  keys = {
+    {
+      "ff", -- try it if you didn't it is a banger keybinding for a picker
+      function()
+	require('fff.main').find_files()
+      end,
+      desc = 'Find files',
+    }
+  }
 }
+```
+
+You can also avoid calling `setup` and simply set `vim.g.fff` instead.
+
+> [!Important]
+> While we are in beta it is required build native backend manually by running `cargo build --release` in the plugin directory.
+
+```lua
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -51,10 +52,10 @@ FFF.nvim requires:
   build = "cargo build --release",
```
jla2000:
```suggestion
  build = "cargo build --release",
  -- Prevent lazy.nvim from lazy loading plugin
  lazy = false,
```

Would it make sense to add this to the suggested plugin spec?

mrcjkb: I don't use lazy.nvim, so I don't know for sure. But I believe adding `keys = ...` makes it lazy-load? Not sure if setting `lazy = false` overrides that? 

I could also put the keymap example in a separate code block that uses `vim.keymap.set`.

jla2000: Hmm I'm also not sure `keys` would make it lazy load again, even when `lazy = false` is specified.

At least for `default.config.lazy` it seems to get overriden by `keys` according to this page: https://lazy.folke.io/spec/lazy_loading

jla2000: Maybe the safe way would really be to set `lazy = false` and then use `vim.keymap.set` in a `config()` function inside the plugin spec

mrcjkb: lazy-loading isn't bad. It's just redundant.
But I do think using `vim.keymap.set` is a better example, as it also works for users of other plugin managers.

mrcjkb: I've added an `init` with `vim.keymap.set`.
Note: I chose `init` over `config`, because if you use `config`, lazy.nvim won't automatically pass the `opts` to the `setup` function.

dmtrKovalenko:
Sorry for the long response—I had a very busy week.

I tested it out, and it appears to be working, although the lazy sync never actually started before the picker is open. Perhaps I'm missing something? 

Also, I (and I believe most lazy users) prefer setting up the plugin using `keys = {}` and `opts = {}` and controlling laziness solely via the `lazy` property. I understand why we need the `lazy_sync` property, but if we're lazy-loading the plugin using lazy.nvim, can't we just set `lazy_sync` to `true` by default so it starts syncing once it's initialized by lazy?

mrcjkb:
No problem :smile: 
I'm not in a rush.

> I tested it out, and it appears to be working, although the lazy sync never actually started before the picker is open. Perhaps I'm missing something?

Should be fixed in the latest commit. I tripped over the `nil`/`false` duality.

> Also, I (and I believe most lazy users) prefer setting up the plugin using keys = {} and opts = {} and controlling laziness solely via the lazy property. 

Okay, I've reverted the lazy.nvim example modification to use the `keys` field again.

> I understand why we need the lazy_sync property, but if we're lazy-loading the plugin using lazy.nvim, can't we just set lazy_sync to true by default so it starts syncing once it's initialized by lazy?

I'm not sure how that would work?
If someone configures lazy.nvim to lazy-load the plugin using a keymap or a command, lazy.nvim will trigger the keymap or command, which will result in `ensure_initialized` being called.

Issue #122, opened 2025-08-19
No syntax highlighter for .ts files


Exrun94:
.ts files do not have syntax highlighter applied 
.tsx are okay

<img width="1517" height="764" alt="Image" src="https://github.com/user-attachments/assets/ce78e2a6-7d0e-4e70-ab80-eb2069d44212" />

oskarrrrrrr:
`vim.filetype.match` seems to be too shy to guess, known issue: https://github.com/neovim/neovim/issues/27265

Probably the easiest solution is to use `plenary`: https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/filetype.lua#L159
They just use a mapping from file extension (or full file name) to file type: https://github.com/nvim-lua/plenary.nvim/blob/master/data/plenary/filetypes/base.lua
I don't love adding a dependency just for that but perhaps it's ok.

PR #125, opened 2025-08-19
fix: better file type detection

oskarrrrrrr:
Fixes missing highlighting for some file types (e.g. typescript or bash).

Fixes: #122


oskarrrrrrr:
Are we okay with the dependency?

dmtrKovalenko:
Actually not sure about it I think for such a simple function maybe we can just copy the code from plenary sources?

oskarrrrrrr:
Copied the minimum amount of code that does what we want. Removed detecting based on shebang and modeline because I don't think that they're too important (also I would have to copy/reimplement more code for handling files).

I still feel like it's a lot for such a simple feature. We could put the file from data folder inline in `filetype.lua` and simplify logic a bit. Should I do that?

dmtrKovalenko:
Can't push to your branch for some reason  please apply this change to pass CI if possible

diff --git a/_typos.toml b/_typos.toml
index 0354a1a..e93f479 100644
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,5 +1,5 @@
 [files]
-extend-exclude = ["/CHANGELOG.md"]
+extend-exclude = ["/CHANGELOG.md", "data/filetypes/base.lua"]
 
 [default.extend-words]
 noice = "noice"

dmtrKovalenko:
closing in favor of #129 I can not push to your branch

Issue #112, opened 2025-08-18
Strange highlighting in prompt buffer


catgoose:
It seems lua keywords are being highlighted?

<img width="280" height="118" alt="Image" src="https://github.com/user-attachments/assets/caab60ac-bba1-417f-bfd5-5bcc2f71a26c" />

<img width="769" height="97" alt="Image" src="https://github.com/user-attachments/assets/eb386019-f392-42ff-b209-599f4282f7d3" />

I'm not sure if this is a problem on my system, but I doubt it.  If I do `<C-c>` on the input then `<esc>` to enter normal mode I can see the filetype is `fff_input` and buftype is `prompt`

PR #127, opened 2025-08-20
fix: disable syntax highlighting for input buffer

oskarrrrrrr:
When `syntax` is enabled Neovim tries really hard to provide syntax highlighting. This is the simplest way I found to disable it for just one buffer.

Setting `syntax = ''` when the buffer is created is not enough, it will get overwritten on changes to the buffer and setting `syntax = ''` on buffer change wasn't 100% reliable for some reason.

Buffer scoped autocmds are cleaned up when the buffer is destroyed.

Fixes: https://github.com/dmtrKovalenko/fff.nvim/issues/112

### DIFF
--- a/lua/fff/picker_ui.lua
+++ b/lua/fff/picker_ui.lua
@@ -437,6 +437,13 @@ function M.setup_buffers()
   vim.api.nvim_buf_set_option(M.state.input_buf, 'filetype', 'fff_input')
   vim.fn.prompt_setprompt(M.state.input_buf, M.state.config.prompt)
 
+  -- Changing the contents of the input buffer will trigger Neovim to guess the language in order to provide
+  -- syntax highlighting. This makes sure that it's always off.
+  vim.api.nvim_create_autocmd('Syntax', {
+    buffer = M.state.input_buf,
+    callback = function() vim.api.nvim_buf_set_option(M.state.input_buf, 'syntax', '') end,
+  })
+
   vim.api.nvim_buf_set_option(M.state.list_buf, 'buftype', 'nofile')
   vim.api.nvim_buf_set_option(M.state.list_buf, 'filetype', 'fff_list')
   vim.api.nvim_buf_set_option(M.state.list_buf, 'modifiable', false)


lua/fff/picker_ui.lua
```diff
@@ -437,6 +437,13 @@ function M.setup_buffers()
   vim.api.nvim_buf_set_option(M.state.input_buf, 'filetype', 'fff_input')
   vim.fn.prompt_setprompt(M.state.input_buf, M.state.config.prompt)
 
+  -- Changing the contents of the input buffer will trigger Neovim to guess the language in order to provide
+  -- syntax highlighting. This makes sure that it's always off.
+  vim.api.nvim_create_autocmd('Syntax', {
+    buffer = M.state.input_buf,
+    callback = function() vim.api.nvim_buf_set_option(M.state.input_buf, 'syntax', '') end,
```
dmtrKovalenko, line 444: why do we need to do this on autocmd if we can set the option directly on a buffer when creating it

dmtrKovalenko:
with your changes I can still see the highlights in the prompt 
image

lua/fff/picker_ui.lua
oskarrrrrrr: Setting it once is not enough. Neovim tries to guess the file type as you type and it will overwrite it.

This is what happens when you set it just once, immediately after opening the search:
Screenshot 2025-08-23 at 17 32 33

and after you type something that looks like Lua:
Screenshot 2025-08-23 at 17 32 52

We can see that Neovim overwrites the value.

oskarrrrrrr:
I just double checked and it works for me. Can you open a new neovim instance, open fff and run `:verbose set syntax?` (you need to press <C-c> first to escape the prompt) and then type "if" in the prompt and run the command again?

dmtrKovalenko:
this is what I get 

```
20:27:29 msg_show.list_cmd   verbose set syntax?   syntax=fff_input
```

while there is still syntax highlighting

dmtrKovalenko:
sorry it's me being dumb

PR #144, opened 2025-09-02
feat: Smart case bonuses

dmtrKovalenko:


mathjiajia:
what happens with `error: Nightly 2025-09-01 is not available`?

dmtrKovalenko:
I didn’t have much time to figure out but seems like nix mirror doesn’t have this nightly yet

jla2000:
@dmtrKovalenko 
If the rust-toolchain is updated, the `rust-overlay` flake input likely needs to be updated as well, this can be done by simply running `$ nix flake update rust-overlay`. It will produce an updated lock file like in https://github.com/dmtrKovalenko/fff.nvim/pull/146

Issue #164, opened 2025-10-02
Error loading module 'blink_cmp_fuzzy'


Irdis:
Hi,
I have both extension `fff` and `blink.cmp`. From time to time I catch the following exception when opening nvim
```
Error executing vim.schedule lua callback: error loading module 'blink_cmp_fuzzy' from file 'C:/Users/XXX/AppData/Local/nvim-data/lazy/fff.nvim/lua/../target/libfff_nvim.dll':
    The specified procedure could not be found.
stack traceback:
    [C]: at 0x7ffe06d7ba50
    [C]: at 0x7ffe06d7bd60
    [C]: in function 'require'
    ...al/nvim-data/lazy/blink.cmp/lua/blink/cmp/fuzzy/init.lua:17: in function 'set_implementation'
    ...ta/Local/nvim-data/lazy/blink.cmp/lua/blink/cmp/init.lua:24: in function 'fn'
    vim/_editor.lua:366: in function <vim/_editor.lua:365>
```
upd.
It is something related to specifying cpath. I tried to print package.cpath and ran couple tests.

If I have the following order I get the error
```
.\?.dll;
C:\tools\neovim\nvim-win64\bin\?.dll;
C:\tools\neovim\nvim-win64\bin\loadall.dll;
C:/Users/User/AppData/Local/nvim-data/lazy/fff.nvim/lua/../target/libfff_nvim.dll;
C:/Users/User/AppData/Local/nvim-data/lazy/fff.nvim/lua/fff/rust/../../../target/release/lib?.dll;
C:/Users/User/AppData/Local/nvim-data/lazy/fff.nvim/lua/fff/rust/../../../target/release/?.dll;
C:/Users/User/AppData/Local/nvim-data/lazy/blink.cmp/lua/blink/cmp/fuzzy/rust/../../../../../target/release/lib?.dll;
C:/Users/User/AppData/Local/nvim-data/lazy/blink.cmp/lua/blink/cmp/fuzzy/rust/../../../../../target/release/?.dll
```

If `blink.cmp` manages to initialize first, I don't see the error and both plugins load correctly. The order is following in this case
```
.\?.dll;
C:\tools\neovim\nvim-win64\bin\?.dll;
C:\tools\neovim\nvim-win64\bin\loadall.dll;
C:/Users/User/AppData/Local/nvim-data/lazy/blink.cmp/lua/blink/cmp/fuzzy/rust/../../../../../target/release/lib?.dll;
[...]

PR #165, opened 2025-10-03
fix: error loading module '...' from file libfff_nvim

Irdis:
It addresses issue https://github.com/dmtrKovalenko/fff.nvim/issues/164

### DIFF
--- a/lua/fff/rust/init.lua
+++ b/lua/fff/rust/init.lua
@@ -7,12 +7,22 @@ local function get_lib_extension()
   return '.so'
 end
 
+---@param bin_path string
+---@return string
+local function get_binary_cpath(bin_path)
+    local bin_dir = vim.fn.fnamemodify(bin_path, ':h')
+    local bin_file = vim.fn.fnamemodify(bin_path, ':t')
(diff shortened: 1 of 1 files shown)


SandyJackson:
Can confirm same error on Mac OS (ARM) is fixed by this change. 

In my case the file ext. is `.dylib` as opposed to `.dll` (as you would expect). 

FWIW `libfff_nvim.dylib` exists in `target/`. But Irdis your explanation in the issue makes sense to my (amateur) eye.

The fix also works if you just go back to `donwnload.lua` and change the `get_binary_path` function (line 19-23).

```
local function get_binary_path(plugin_dir)
  local binary_dir = get_binary_dir(plugin_dir)
  local extension = system.get_lib_extension()
  return binary_dir .. '/lib?.' .. extension
end
```

I am no expert and cannot comment on if this is the best/an appropriate solution but the above change is functionally the same and simpler than the current PR. (great work finding and fixing though @Irdis you've saved me hours of work that probably still wouldn't have yielded a result!

EDIT: I also use blink.cmp

PR #166, opened 2025-10-04
fix cpath related errors

madmaxieee:
Current way of configuring cpath use the full path of the downloaded binary without any '?', this causes any require statement that can't resolve before this cpath component to always resolve to the downloaded libfff_nvim, thus causing some unexpected bug in some other plugins or user code.

fixes errors like this:
```
Error executing vim.schedule lua callback: error loading module 'blink_cmp_fuzzy' from file '/Users/quolpr/.local/share/nvim/lazy/fff.nvim/lua/../target/libfff_nvim.dylib':
        dlsym(0x92987180, luaopen_blink_cmp_fuzzy): symbol not found
stack traceback:
        [C]: at 0x0103251790
        [C]: at 0x0103251b88
        [C]: in function 'require'
        ...l/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/init.lua:17: in function 'set_implementation'
        .../.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/init.lua:24: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

### DIFF
--- a/doc/fff.nvim.txt
+++ b/doc/fff.nvim.txt
@@ -1,4 +1,4 @@
-*fff.nvim.txt*       For Neovim >= 0.10.0       Last change: 2025 September 17
+*fff.nvim.txt*        For Neovim >= 0.10.0        Last change: 2025 October 04
 
 ==============================================================================
 Table of Contents                                 *fff.nvim-table-of-contents*
--- a/lua/fff/download.lua
+++ b/lua/fff/download.lua
@@ -187,4 +187,11 @@ function M.get_binary_path()
   return get_binary_path(plugin_dir)
 end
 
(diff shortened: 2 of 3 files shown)


SandyJackson:
This addresses issue #164 and is a different, but similar approach to PR #165

Just to keep everything linked 👍

madmaxieee:
ah, right, I didn't see that PR, thanks

dmtrKovalenko:
I'll  keep the other PR, thank you both for contributing. And sorry for the delay

dmtrKovalenko:
No I actually will keep this one because I can not update that PR cause github sucks

PR #176, opened 2025-10-20
fix: Crash on ts filetype if plenary is not installed

dmtrKovalenko:
(dumb regression from the previous PR)


amadeus:
Appears to work great, thanks for the fix!

PR #183, opened 2025-12-06
add default support for vertical navigation with ctrl j-k

marcosktsz:


dmtrKovalenko:
I don't think this is a standard. We by default support <C-n> and <C-p> so feel free to add those to your own config

Issue #160, opened 2025-09-18
Space is treated as part of the search query


jla2000:
When searching for a specific file, I often input a few words that are located in the files path.
However, it seems like the behavior recently changed and space is now treated as part of the search query.
This means it's no longer possible to find a file using some keywords that are separated by space.

LeoPatOZ:
Any update on this?

PR #184, opened 2025-12-06
Space is treated as part of the search query

builtbyleo:
Hopefully fixes https://github.com/dmtrKovalenko/fff.nvim/issues/160

I think the easiest way to solves this issue is to simply consider spaces between words as the user specifying a filepath. 

So now "helper mod" matches to helper/mod etc.. (each search term is still typo resistant etc).

Also added whitespace trimming to begging of search query so "    mod" will search for "mod". 

Let me know if you think this is the right approach (i tried to go with the simplest / minimal change route)

### DIFF
--- a/lua/fff/rust/score.rs
+++ b/lua/fff/rust/score.rs
@@ -522,4 +522,88 @@ mod tests {
         assert_eq!(items[1].relative_path, "file6.rs");
         assert_eq!(items[2].relative_path, "file4.rs");
     }
+
+    #[test]
+    fn test_spaces_become_path_separators() {
+        // Files with nested paths
+        let files = vec![
+            FileItem {
+                path: PathBuf::from("src/helper/mod.rs"),
+                relative_path: "src/helper/mod.rs".to_string(),
+                relative_path_lower: "src/helper/mod.rs".to_string(),
+                file_name: "mod.rs".to_string(),
+                file_name_lower: "mod.rs".to_string(),
+                size: 0,
+                modified: 1000,
+                access_frecency_score: 0,
+                modification_frecency_score: 0,
+                total_frecency_score: 0,
+                git_status: None,
+            },
+            FileItem {
+                path: PathBuf::from("src/helpermod.rs"),
+                relative_path: "src/helpermod.rs".to_string(),
+                relative_path_lower: "src/helpermod.rs".to_string(),
+                file_name: "helpermod.rs".to_string(),
+                file_name_lower: "helpermod.rs".to_string(),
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
Thank you but I am not sure that what people actually need is the matching of "helper mod" to match on helper/mod but instead they want matcher to match all the results of helper and then run the search of mod within a previous search

builtbyleo:
Thank you for the feedback!

Funnily enough I initially implemented it this way but found that I was inadvertently searching for files within a path hence this approach.

Maybe I'm slightly confused but do you mind giving me an example where the current approach matches on a query when it shouldn't ?

(I guess im asking if you think this approach would match too much or too little)

(I was basing the implementation based of this issue https://github.com/dmtrKovalenko/fff.nvim/issues/160)

Thanks again for the review!

dmtrKovalenko:
I am pretty sure that for the path like

'rust/helper/mod.rs' people are expecting to put query like "helper mod rust" and get a match which is reasonable and likely useful. While your implementation requires to put rust helper mod which is weird because if you know the path you can always match it using slashes

isentropic:
You mentioned that you work in monorepos, and this exactly my issue. For example I remember that the file is somewhere

`service1/..../lib.../java....test/....buy.java`, 

but there are others like 
`src /..../app.../java....test/....buy.java`  
`service2 /..../lib.../java....test/....buy.kt`


in this case I would say `service1 lib java test buy`, and this would very true in all big monorepos. Somehow you remember only middle parts like `src` or `lib` or `app` or `test` and the service name like `backend` and `fronend` in front so that your query would be `src app backend something`

dmtrKovalenko:
Yeah this is exactly something I want to do, I am working to make this possible but this requires quite redesign

isentropic:
amazing, and amazing work overall. Now that I started working in monorepos all other picker plugins are just plain unsatisfactory. Somehow VSCode and other normie editors get this right (or at least better than just `find`) out of the box. I also wonder how they do things in their implementation.

as a next idea  you could also somehow rank using frequency, history, relative locality, exact matching score, and let's say commits on a file e.t.c. build some sort of recommendation/search engine

dmtrKovalenko:
Your last paragraph just describes the reason fff was done, no? We have all of that since the very start

builtbyleo:
> Yeah this is exactly something I want to do, I am working to make this possible but this requires quite redesign

I think given this + discussion on this thread I think its best if i close this PR and wait until someone who actually knows what they're doing takes a crack at it

PR #187, opened 2025-12-08
feat: Last Query Match

### LINKED ISSUE #152: Preserve prompt for a customizable time after making a selection
I think it would be useful to keep prompt and search result for some time, e.g. 60 sec. 

E.g. you typed something and found multiple somewhat similar files, e.g. files with the same name but in different namespaces so you select the first one but it's wrong so you need to go back and type the same prompt again. Instead of doing that it would be cool to resume previous state of the most recent search within septic time period.

### LINKED ISSUE #159: Ability to resume last search
Similar to `Snacks.picker.resume()` which restores the last search input when opening up the search window. This is useful to prevent typing the same search input over and over again when you want to jump between different files.

https://github.com/folke/snacks.nvim/blob/main/docs/picker.md#resume

dmtrKovalenko:
closes #159
closes https://github.com/dmtrKovalenko/fff.nvim/issues/152


justDeeevin:
How can I use this feature to do what is mentioned in #159? Is this documented anywhere? I can't seem to find anything.

dmtrKovalenko:
@justDeeevin there is a built in feature for this `<C-Up>`

justDeeevin:
is this a bind while you're in a picker? if so, it's not quite what i'm looking for. I would like to be able to have a bind to open my last-opened search (be it grep or file search)

dmtrKovalenko:
It's pretty easy to add as well, PR is welcome

dmtrKovalenko:
Or you can try to open a picker and then feed the <C-Up> manually as well. But PR is more appreciated

justDeeevin:
See #371 :>

PR #248, opened 2026-02-19
fix: Use consistent emoji for frecency debug star indicator

koichiio:
Change the star from (U+2B50) to (U+2B50 + U+FE0F) to enforce it to be an emoji for the debugging frequency mode (You can see the text-based star in the screenshot below). It seemed off and once tried to make the frecency icon stuff configurable, but just realized it's just a debug feature. Very nit-picky change, so feel free to skip it if you don't like. 

2026-02-18-163602_hyprshot

### DIFF
--- a/lua/fff/file_renderer.lua
+++ b/lua/fff/file_renderer.lua
@@ -72,7 +72,7 @@ function M.render_line(item, ctx, item_idx)
       if mod >= 6 then
         indicator = '🔥'
       elseif access >= 4 then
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
Need to make sure that it doesn't break layouting because emojies because I might have used somewhere the byte length instead of vim.strdisplaywidth

dmtrKovalenko:
seems to be fine

Issue #249, opened 2026-02-19
Flex layout


oyarsa:
Telescope has a flex layout:

```lua
require("telescope").setup({
    defaults = {
        layout_strategy = "flex",
        layout_config = {
            flip_columns = 130,
        },
    },
})
```

Where it flips from showing the preview on top or the side, depending on the screen width. Would it be possible to add something like this?

dmtrKovalenko:
So you demand responsive layout that wraps the layout based on the size of the screen?

oyarsa:
Yes, that's the idea. In the example above, if the window is less than 130 columns wide, the layout places the preview on top. If more than that, then the preview is placed on the right.

PR #255, opened 2026-02-20
feat: Add preview_position = 'flex'

oyarsa:
closes https://github.com/dmtrKovalenko/fff.nvim/issues/249

As mentioned in #249, this PR introduces a new preview position value, `flex` and an associated option `flex_columns`. If the `terminal_width` (`vim.o.columns`) is less than `flex_columns`, the preview is placed on `top`. Otherwise, it's placed on `right`.

The default value of 130 was used because it's roughly half of what I get on a full-screen window on 1080p.

Note: this also auto-evaluate if the window is resized

### DIFF
--- a/lua/fff/picker_ui.lua
+++ b/lua/fff/picker_ui.lua
@@ -89,25 +89,192 @@ local function get_preview_position()
   return 'right'
 end
 
---- Function-based config options:
---- config.layout.width: number|function(terminal_width, terminal_height): number
---- config.layout.height: number|function(terminal_width, terminal_height): number
---- config.layout.preview_size: number|function(terminal_width, terminal_height): number
---- config.layout.preview_position: string|function(terminal_width, terminal_height): string
---- config.layout.prompt_position: string|function(terminal_width, terminal_height): string
+local function compute_layout(config)
+  local debug_enabled_in_preview = M.enabled_preview() and config.debug and config.debug.show_file_info or false
 
---- @class LayoutConfig
---- @field total_width number
---- @field total_height number
---- @field start_col number
---- @field start_row number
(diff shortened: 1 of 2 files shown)


lua/fff/picker_ui.lua
```diff
@@ -60,21 +60,31 @@ local function get_prompt_position()
...
-      function(value) return utils.is_one_of(value, { 'left', 'right', 'top', 'bottom' }) end,
+      function(value) 
+        return utils.is_one_of(value, { 'left', 'right', 'top', 'bottom', 'flex' }) 
+      end,
       'right',
       'layout.preview_position'
     )
+
+    if position == 'flex' then
+      return terminal_width >= config.layout.flex_columns and 'right' or 'top'
```
dmtrKovalenko: Someone would likely want to be able to put the preview underneath wdyt? 

what about making flex as an object something like 

preview_position = 'right'
flex = {
    size = 130,
    wrap = 'top'
}

this will not have a breaking change and I think we can even have it as a default

lua/fff/picker_ui.lua
```diff
@@ -60,21 +60,31 @@ local function get_prompt_position()
   return 'bottom'
 end
 
+local DEFAULT_FLIP_COLUMNS = 130
+
 local function get_preview_position()
   local config = M.state.config
 
   if config and config.layout and config.layout.preview_position then
     local terminal_width = vim.o.columns
```
dmtrKovalenko, line 67: I bet this is not getting rerendered on resizing neovim it would be cool to add autogroup for this so it is resizing on the flight

oyarsa:
@dmtrKovalenko following your flex is now the default, set to `nil` to disable. I'm not quite confident about this, though. Let me know what you think.

I'll try to implement the auto-resize next, but that seemed a little more involved.

PR #288, opened 2026-03-16
feat(ui): config option to override buffer switching logic

### LINKED ISSUE #287: [Discussion] unintuitive `find_suitable_window` logic for `edit` actions
## Problem

Quite often, I find myself launching the fff.nvim picker from a terminal buffer. When I select a file with `<CR>`, it opens the buffer in anther split that I have open.

I can see how this could be useful to prevent opening files in buffers like the quickfix list.
But more often than not, I actually want it to replace my buffer, even if it's not a "normal" buffer.
i.e., I'll end up with the following situation:

```
<regular_buffer> | <terminal_buffer>
...              |
...              | > | <- cursor is here
```

I want to open a file and replace the terminal buffer.
With the current implementation, I have to:

- Close the terminal buffer
- Create a new split and search the file I want to replace the terminal buffer with
- Or, Search and use the keymap for opening in a vertical split

Neither is intuitive (and it's inconsistent with the behaviour of other pickers, like telescope.nvim).

I have the same issue with other buffers, like oil.nvim buffers (buftype = "acwrite").

## Proposed solutions

- Remove the logic altogether
- Or, add a config option to disable it
- Or, maintain a list of buffer types (e.g. `quickfix`) that trigger the `find_suitable_buffer` switching logic.
- Or, maintain a list of suitable buffer types (e.g. `terminal` and `acwrite`)

## Relevant code

https://github.com/dmtrKovalenko/fff.nvim/blob/64861f8142c5ce8c34f9ec73a3d084857182e156/lua/fff/picker_ui.lua?plain=1#L2305-L2308

mrcjkb:
This is my suggestion to fix #287.


dmtrKovalenko:
I'd probably have a suggestion to have a function override here we might want to extend the logic with more heuristics there for example if the buffer is modifiable or not

mrcjkb:
> for example if the buffer is modifiable or not

I removed that check because it seemed too strict to me (terminal buffers aren't modifiable, for instance).

dmtrKovalenko:
Please do not remove any checks. There are a plenty of buffers of special file types that we should never replace

this is simply the worst experience when your file opens in the sidebar which is always happens to snacks nvim for example

mrcjkb:
> Please do not remove any checks. There are a plenty of buffers of special file types that we should never replace
> 
> this is simply the worst experience when your file opens in the sidebar which is always happens to snacks nvim for example

Do you have an example?
I don't think I use any plugins that create a "sidebar", which would explain why I haven't experienced any issues without the modifiable check.

dmtrKovalenko:
https://github.com/hedyhli/outline.nvim I use this one personally

mrcjkb:
Hmm, how about keeping the behaviour as is and adding a config option to disable it altogether? Or a config option to override the behaviour with a function?
I don't use sidebar or file tree plugins, so for me, having my selection replace a seemingly arbitrary buffer isn't great.

dmtrKovalenko:
Yeah that’s what I’m saying you can make a function that overrides the existing one without changing the behavior

mrcjkb:
Done. I moved the logic for finding a suitable window to the config module. That might not be the best place - perhaps you'd want a hooks module or something.

lua/fff/conf.lua
```diff
@@ -309,6 +351,21 @@ local function init()
       min_combo_count = 3, -- Minimum selections before combo boost applies (3 = boost starts on 3rd selection)
       combo_boost_score_multiplier = 100, -- Score multiplier for combo matches (files repeatedly opened with same query)
     },
+    select = {
+      --- @param current_buf integer
+      --- @param action FffSelectAction
+      pre_select_hook = function(current_buf, action)
```
dmtrKovalenko, line 357: I don't know if this is actually a hook

dmtrKovalenko (approved):
thx!

lua/fff/conf.lua
mrcjkb: Shall we call it `pre_select_callback` or `on_pre_select`?

dmtrKovalenko: I mean it's unclear that this hook should actually change the window, maybe we should just override the whole open_file callback that is doing all the window picking and :e call 

this would be more flixble probably

dmtrKovalenko: or the other option is to make this callback return the window_id and we then opening the file in this window

mrcjkb: Hmm, I like the idea of returning a window ID (which I guess would make it a `select_window_hook`).

mrcjkb:
Closing, as I no longer use fff.nvim.

dmtrKovalenko:
😭😭😭😭 why?

mrcjkb:
> 😭😭😭😭 why?

Gave it a try for a while, but it has some quirks that don't fit into my workflow. Most importantly:

- It sometimes lists recently deleted or moved files with high priority (possibly because the in-memory index needs time to catch up + high recency score).
- Files that don't match what I'm looking for with a higher frequency/recency score often get listed first because of the typo-resistant fuzzy matching algorithm. I run into this more often than I mistype.
- It ignores files in hidden directories that are checked into scm.
- With telescope, I mostly use `jj file list`, which has had good performance in all of the projects I work with.

Keep up the good work 😃
I'm sure fff.nvim is great for many people, it's just not for me.

PR #327, opened 2026-04-03
fix: accept whole-number float maxResults values

### LINKED ISSUE #326: mcp: accept whole-number float maxResults values from clients
## Summary
The MCP server currently rejects `maxResults` when some clients serialize it as an integral float like `30.0`.

## Reproduction
A request like this fails today:

```json
{
  "patterns": ["AgentTool"],
  "constraints": "*.{ts,tsx} src/js/my-app/",
  "maxResults": 30.0
}
```

## Actual
Deserialization fails with:

```text
invalid type: floating point `30.0`, expected usize
```

## Expected
Whole-number floats from MCP clients should be accepted as the equivalent integer value, while fractional values like `30.5` should still be rejected.

## Why this matters
A few MCP clients / AI tool wrappers emit JSON numbers as floats even when the logical value is integral. Being strict here creates avoidable interoperability failures for an otherwise safe parameter.

kuishou68:
Closes #326

## Summary
- accept whole-number float values like `30.0` for `maxResults`
- keep rejecting fractional values like `30.5`
- add serde regression tests for integer, whole-float, and fractional-float cases

## Notes
This is aimed at MCP client interoperability: some wrappers serialize JSON numbers as floats even when the logical value is integral.

## Testing
- not run in this sandbox (`cargo`/`rustc` are unavailable here)
- added focused deserialization tests in `crates/fff-mcp/src/server.rs`

### DIFF
--- a/crates/fff-mcp/src/server.rs
+++ b/crates/fff-mcp/src/server.rs
@@ -70,12 +70,55 @@ fn make_grep_options(
     )
 }
 
+fn deserialize_optional_usize_lenient<'de, D>(deserializer: D) -> Result<Option<usize>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    use serde::de;
+
+    struct OptionalUsizeVisitor;
+
+    impl<'de> de::Visitor<'de> for OptionalUsizeVisitor {
+        type Value = Option<usize>;
+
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
heavily AI generated PR. Fixed myself here https://github.com/dmtrKovalenko/fff.nvim/pull/328

PR #330, opened 2026-04-03
feat: add serialization accessors to BigramFilter

magnusmalm:
## Summary

Adds read-only accessors and a `from_raw_parts` constructor to `BigramFilter`, enabling external tools to serialize and deserialize the bigram index to/from disk without reaching into private fields.

**New public methods:**
- `lookup()`, `dense_data()`, `words()`, `dense_count()`, `populated()` — field accessors
- `skip_index() -> Option<&BigramFilter>` — sub-index reference
- `from_raw_parts(lookup, dense_data, ...) -> Self` — reconstruction from serialized data

This is a purely additive, non-breaking change — no existing code is modified, no new dependencies. All 7 methods are simple one-line accessors/constructors.

## Use case

[fff-cli](https://github.com/magnusmalm/fff-cli) is a standalone CLI built on `fff-core` that persists the bigram index to disk (`.fff/bigram.bin`) for fast startup. It needs to read the filter's raw data for serialization and reconstruct it on load. Without these accessors, the only alternative is rebuilding the bigram index from scratch on every invocation (~200ms), negating the persistent index advantage.

## Test plan

- [ ] Existing tests pass (no code changed, only additions)
- [ ] Verified via fff-cli: index → serialize → deserialize → grep produces correct results on buildroot (13k files)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ PLENARY_DIR ?= ../plenary.nvim
 
 .PHONY: build test test-rust test-lua test-version test-bun test-node prepare-bun prepare-node set-npm-version header
 
+all: format test lint
+
 build:
 	cargo build --release --features zlob
 
--- a/crates/fff-nvim/src/bin/bench_grep_query.rs
+++ b/crates/fff-nvim/src/bin/bench_grep_query.rs
@@ -1,4 +1,3 @@
-use fff::BigramIndexBuilder;
 /// Single-query grep benchmark with bigram index profiling.
 ///
 /// Usage:
@@ -18,7 +17,6 @@ fn fmt_dur(us: u128) -> String {
         format!("{}µs", us)
     }
 }
-
 
 fn run_grep(files: &[fff::FileItem], index: Option<&fff::BigramFilter>, query: &str, iters: usize) {
     let options = GrepSearchOptions {


crates/fff-core/src/bigram_filter.rs
```diff
@@ -365,6 +365,58 @@ impl BigramFilter {
     pub fn has_key(&self, key: u16) -> bool {
         self.lookup[key as usize] != NO_COLUMN
     }
+
+    // ── Serialization accessors ────────────────────────────────────────
```
dmtrKovalenko: please remove this it is screaming OPUS 4.6

crates/fff-core/src/bigram_filter.rs
```diff
@@ -365,6 +365,58 @@ impl BigramFilter {
...
+        self.populated
+    }
+
+    /// Reference to the optional skip-1 bigram sub-index.
+    pub fn skip_index(&self) -> Option<&BigramFilter> {
+        self.skip_index.as_deref()
+    }
+
+    /// Reconstruct a `BigramFilter` from raw parts (for deserialization).
+    pub fn from_raw_parts(
```
dmtrKovalenko: from_raw_parts sounds like it is going to use the from_raw_parts_mut to get the data from the array of bytes. 

To be honest it kinda makes sense, I am not sure how do you want to serialize it but to keep it fast I suggest using something like https://docs.rs/zerocopy/latest/zerocopy/ and make sure that the ABI is not changing by using repr(C) thus you will be able to simply memory map the index in the memory and provide it right away to every call

crates/fff-core/src/bigram_filter.rs
magnusmalm: Removed

crates/fff-core/src/bigram_filter.rs
magnusmalm: The current approach (reconstruct + accessors) was mostly a minimal starting point, that did not walk all over your code. :) I can work on a zerocopy variant. A bit more involved and invasive but I'll give it a go and see what you think.

dmtrKovalenko: I am wondering how are you planning to maintain the index after the changes?

magnusmalm: For the bigram index specifically, the current on-disk format is already pretty close to zerocopy, flat lookup table + dense bitset columns, loaded via mmap. The actual copy into owned Vecs is around 0.2ms so it's not really the bottleneck.

The expensive part is reconstructing ~13k FileItem structs from the file list (~8ms of ~11ms total load time). That's where zerocopy would really pay off, but FileItem owns its strings via String/PathBuf so going full zerocopy on it would be pretty invasive.

What I think could work without too much disruption: a read-only BigramView<'a> that borrows lookup/dense slices directly from an mmap, paired with a trait that both BigramFilter and BigramView implement for the query path. Building still uses the owned BigramFilter, querying can use either. No lifetime changes needed on the existing types.

I could code that up in a new PR and bench it with fff-cli.

dmtrKovalenko: I think that you should persist the file list as well somehow likely in the separate file and lookup indexes

magnusmalm: Yes, agreed. The file list reconstruction is 8ms vs 0.2ms for the bigram. fff-cli already persist it in a separate file (files.bin) with a string table + fixed-size record array, but it still copies into owned FileItem structs on load.

A zerocopy file list would need the record array to be repr(C) and directly mmappable, with paths referenced as offsets into the string table rather than owned Strings. I'll sketch out some code and include it in the BigramView PR.

PR #334, opened 2026-04-04
fix: suspend/restore `set paste` when loading picker UI

anmonteiro:
sometimes I'll have `set paste` on because i'm pasting something from an external clipboard, but it doesn't make sense to have it in the file picker UI.

### DIFF
--- a/lua/fff/picker_ui.lua
+++ b/lua/fff/picker_ui.lua
@@ -442,6 +442,16 @@ end
 local preview_config = conf.get().preview
 if preview_config then preview.setup(preview_config) end
 
+local function suspend_paste()
+  if not vim.o.paste then return false end
+  vim.o.paste = false
+  return true
+end
+
+local function restore_paste(should_restore)
+  if should_restore then vim.o.paste = true end
+end
+
 M.state = {
   active = false,
   layout = nil,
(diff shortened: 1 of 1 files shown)


lua/fff/picker_ui.lua
```diff
@@ -442,6 +442,16 @@ end
 local preview_config = conf.get().preview
 if preview_config then preview.setup(preview_config) end
 
+local function suspend_paste()
```
dmtrKovalenko, line 445: Why not to override it in the buffer local options? Doesn’t it work?

anmonteiro:
`:set paste` is actually a global setting, not per buffer.

dmtrKovalenko:
You can override any global setting per buffer using vim.bo, we do that for a bunch of settings 

It would be a more neat solution

anmonteiro:
wasn't aware, let me try!

anmonteiro:
that doesn't seem to be possibile. 

This PR is mostly a nice-to-have, and not strictly needed for me. was just jarring when my vim was in paste mode and navigating in the fff picker didn't work.

dmtrKovalenko:
I'll check and if it's not I'll merge it

PR #335, opened 2026-04-04
fix(preview): avoid ftplugin side effects in scratch preview buffers

anmonteiro:
noticed that e.g. my Merlin plugin was being loaded while cycling through file results in the picker.

this does what telescope does, it avoids running the filetype events and just loads the filetype highlighter.

### DIFF
--- a/lua/fff/file_picker/preview.lua
+++ b/lua/fff/file_picker/preview.lua
@@ -6,6 +6,28 @@ local rust = require('fff.rust')
 
 local M = {}
 
+-- Preview buffers are scratch buffers. Detect the file's language and attach
+-- highlighting directly, but keep buffer filetype empty to avoid ftplugin and
+-- LSP side effects that are meant for real editing buffers.
+local function attach_preview_highlighter(bufnr, filetype)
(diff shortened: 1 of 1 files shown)


dmtrKovalenko (approved):

mrcjkb:
I don't think this is a good change.

Reason: This forces preview highlighting on the user.
Neovim's philosophy allows the user to configure syntax highlighting on ftplugin events. For example, I may want to conditionally enable tree-sitter highlighting depending on the file size in languages with slow parsers or inefficient queries. **With tree-sitter highlighting enabled unconditionally, preview scratch buffers can really slow down the UI.**

Before this change, fff.nvim already set all the right properties to prevent side-effects like starting LSP clients.

If this is about lazy-loading plugins:
See [`:h lua-plugin-lazy`](https://neovim.io/doc/user/lua-plugin/#lua-plugin-lazy).
If a plugin is misbehaving in a scratch buffer because it emits a filetype event, that should be fixed in the plugin (e.g. it can check the `buftype` property).

There is nothing in [`:h filetype`](https://neovim.io/doc/user/filetype/#file-types) that states that scratch/preview buffers cannot have a filetype. Another example of buffers that have filetypes are fugitive.vim diffs.

PR #336, opened 2026-04-04
fix: implement clear_cache function for FFFClearCache command

### LINKED ISSUE #169: `FFFClearCache` user command: function `clear_cache` does not exist
There currently is an `:FFFClearCache` user command, which invokes `require "fff".clear_cache()`, which does not exist.
This would probably need to be added to the module in main.lua?

https://github.com/dmtrKovalenko/fff.nvim/blob/2df06289c9a20545d9d060ceb26cc0cffd53ab9e/plugin/fff.lua#L60-L64

JasonOA888:
## Summary

The `:FFFClearCache` user command calls `require("fff").clear_cache()`, but this function did not exist, causing an error when running the command.

## Changes

- Add `clear_cache(scope)` function to `lua/fff/main.lua`
- Supports three modes:
  - `"all"` (default) - clears both frecency and history caches
  - `"frecency"` - clears only frecency database
  - `"files"` - clears only query history database

## Testing

```vim
:FFFClearCache        " clears all caches
:FFFClearCache all    " same as above
:FFFClearCache frecency
:FFFClearCache files
```

Closes #169

### DIFF
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -216,10 +216,13 @@ function M.clear_cache(scope)
   scope = scope or 'all'
 
   local config = require('fff.conf').get()
+  local fuzzy = require('fff.fuzzy')
   local cleared = {}
 
   -- Clear frecency database
   if scope == 'all' or scope == 'frecency' then
+    -- Properly close LMDB before deleting to avoid memory-mapped file issues
+    pcall(fuzzy.destroy_db)
     local db_path = config.frecency and config.frecency.db_path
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
Have you actually tested it? lmdb is mapping the files and this can crush neovim and likely wouldn’t work on windows (which is fine tbh)

dmtrKovalenko:
closed by https://github.com/dmtrKovalenko/fff.nvim/commit/ea1f9802d7879345f02bef72bcadecd7252b8add

PR #338, opened 2026-04-04
fix: restore FFFClearCache command implementation

### LINKED ISSUE #169: `FFFClearCache` user command: function `clear_cache` does not exist

kuishou68:
## Summary
- add the missing `require('fff').clear_cache()` implementation behind `:FFFClearCache`
- wire the advertised `all|frecency|files` modes to the existing cache cleanup primitives
- add focused Lua tests covering the default, `frecency`, and `files` paths

## Notes
- I couldn't run the Neovim/Rust test stack in this sandbox because `nvim` and `cargo` are unavailable here.
- The fix is intentionally narrow: it restores the missing command target and keeps the cache clearing behavior in one Lua entrypoint.

Closes #169

### DIFF
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -55,6 +55,32 @@ function M.find_in_git_root()
   M.find_files_in_dir(git_root)
 end
 
+--- Clear FFF caches
+--- @param scope? string Cache scope: all|frecency|files
+function M.clear_cache(scope)
+  local fuzzy = require('fff.fuzzy')
+  scope = scope or 'all'
+
+  local ok = true
+
+  if scope == 'all' or scope == 'files' then
+    ok = pcall(fuzzy.cleanup_file_picker)
+    if ok then ok = pcall(fuzzy.destroy_db) end
+  end
+
+  if ok and (scope == 'all' or scope == 'frecency') then
+    ok = pcall(fuzzy.destroy_query_db)
+  end
+
+  if not ok then
+    vim.notify('Failed to clear FFF cache', vim.log.levels.ERROR)
+    return false
+  end
+
+  vim.notify('Cleared FFF cache: ' .. scope, vim.log.levels.INFO)
+  return true
+end
+
 --- Trigger rescan of files in the current directory
 function M.scan_files()
   local fuzzy = require('fff.core').ensure_initialized()
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -55,25 +55,29 @@ function M.find_in_git_root()
   M.find_files_in_dir(git_root)
 end
 
---- Clear FFF caches
+--- Clear FFF caches (both in-memory state and on-disk database files)
 --- @param scope? string Cache scope: all|frecency|files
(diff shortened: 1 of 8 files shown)


lua/fff/main.lua
```diff
@@ -55,6 +55,32 @@ function M.find_in_git_root()
...
 
+--- Clear FFF caches
+--- @param scope? string Cache scope: all|frecency|files
+function M.clear_cache(scope)
+  local fuzzy = require('fff.fuzzy')
+  scope = scope or 'all'
+
+  local ok = true
+
+  if scope == 'all' or scope == 'files' then
```
dmtrKovalenko, line 66: I don't really know if this is something people will be actually using. It just cleans in memory

tests/clear_cache_spec.lua
```diff
@@ -0,0 +1,46 @@
+---@diagnostic disable: undefined-global
+
+describe('clear_cache', function()
```
dmtrKovalenko, line 4: idk about this test it is too mocky but I don't want to make the real test that cleans my db. We should either have a test that provides custom paths for databases and verifies they are cleaned or just get rid from this test

dmtrKovalenko (changes requested):

kuishou68:
I've pushed a fix for the `lua-language-server` type-check failure.

The test constructs minimal `vim.g.fff` configs (only `frecency`/`history` fields) which triggered `missing-fields` warnings for the `FffConfig` and `FffHistoryConfig` types. Since the tests intentionally use partial configs, I've added `missing-fields` to the existing `---@diagnostic disable` directive at the top of the test file.

This keeps the type safety guarantees for production code while allowing the test harness to inject the minimal configuration it needs. Could you clarify what specific changes you'd like for the implementation in the main source files? Happy to iterate.

kuishou68:
Thanks for the feedback! Addressing both inline comments:

**On `lua/fff/main.lua:66` — practical use of `clear_cache`:**

The primary use case is users wanting to reset their frecency ranking (e.g. after a large project restructure, or after switching branches) without restarting Neovim. The `files` scope specifically handles clearing the in-memory file picker index and triggering a fresh scan — useful when `fff` is picking up stale entries. If you feel the scope is too niche, I'm happy to narrow it or remove the function entirely.

**On `tests/clear_cache_spec.lua:4` — test approach:**

The test already uses the approach you described: `vim.fn.tempname()` creates real temporary paths on disk, `fff_rust.init_db()` creates actual LMDB directories there, and the test verifies those directories are deleted after `clear_cache()`. No mocking of the filesystem or Rust layer. The `after_each` cleanup removes leftover temp files so it never touches the user's real databases.

Would this address your concerns, or would you prefer a different structure?

PR #343, opened 2026-04-06
fix(grep): align preview highlight stripping with Rust query parser

### LINKED ISSUE #331: grep preview highlight logic diverges from core query parser semantics
## Summary

The grep preview highlight logic in `lua/fff/location_utils.lua` appears to re-parse the query using a simplified Lua heuristic instead of using the same parsed grep text semantics as the Rust query parser.

As a result, preview highlighting can diverge from the actual grep query behavior when the query contains multiple text tokens, constraints, or escaped constraint-like tokens.

## Current Behavior

In `lua/fff/location_utils.lua`, the preview highlight code derives `search_text` like this:

```lua
local parts = vim.split(query, '%s+')
local text_parts = {}
for _, part in ipairs(parts) do
  if part ~= '' and not part:match('^[%*!/]') and not part:match('^%.') then
    table.insert(text_parts, part)
  end
end
if #text_parts > 0 then search_text = text_parts[1] end
```

This means the preview highlighter effectively uses only the first non-constraint-looking token.

But the Rust parser builds grep text differently. In `crates/fff-query-parser/src/parser.rs`:

```rust
pub fn grep_text(&self) -> String {
    match &self.fuzzy_query {
        FuzzyQuery::Empty => String::new(),
        FuzzyQuery::Text(t) => strip_leading_backslash(t).to_string(),
        FuzzyQuery::Parts(parts) => parts
            .iter()
            .map(|t| strip_leading_backslash(t))
            .collect::<Vec<_>>()
            .join(" "),
    }
}
```

So the effective grep text is based on the parsed non-constraint text parts, joined back together, with escaped constraint tokens handled according to parser rules.

## Why this is a problem

The preview UI and the core parser can disagree about what the actual search text is.

That can produce incorrect highlight behavior for queries such as:

- `*.rs TODO fixme`
- `/src/ init config`
- `!test foo bar`
- `\\*.rs literal`

In those cases, the result set may be produced using the parser's real grep text semantics, while preview highlighting only tracks the first remaining token (for example just `TODO` instead of `TODO fixme`).

## Expected Behavior

Preview highlighting should use the same parsed grep text semantics as the core query parser, rather than re-implementing a simplified constraint-strip heuristic in Lua.

That would keep the preview UI aligned with actual grep matching behavior.

### Preview heuristic in Lua

`lua/fff/location_utils.lua`:

```lua
local parts = vim.split(query, '%s+')
local text_parts = {}
for _, part in ipairs(parts) do
  if part ~= '' and not part:match('^[%*!/]') and not part:match('^%.') then
    table.insert(text_parts, part)
  end
end
if #text_parts > 0 then search_text = text_parts[1] end
```

### Core grep text semantics in Rust

`crates/fff-query-parser/src/parser.rs`:

```rust
pub fn grep_text(&self) -> String {
    match &self.fuzzy_query {
        FuzzyQuery::Empty => String::new(),
        FuzzyQuery::Text(t) => strip_leading_backslash(t).to_string(),
        FuzzyQuery::Parts(parts) => parts
            .iter()
            .map(|t| strip_leading_backslash(t))
            .collect::<Vec<_>>()
            .join(" "),
    }
}
```

## Reproduction

1. Open a grep result in preview mode.
2. Use a query that contains both constraints and multiple text tokens, for example:
   - `*.rs TODO fixme`
   - `/src/ init config`
   - `!test foo bar`
3. Compare:
   - the actual result selection behavior
   - the text highlighted in the preview
4. Observe that the preview highlight logic appears to be driven by a simplified first-token heuristic instead of the parser's effective grep text.

## Duplicate Check

- Open issues checked: no direct match found
- Closed issues checked: no direct match found
- Recent PRs checked: no direct match found

mvanhorn:
The Lua heuristic in `highlight_grep_matches` (location_utils.lua) used a simple prefix check to strip constraint tokens before highlighting. This diverged from the Rust `GrepConfig` parser in several ways:

- Multi-word queries like `foo bar *.rs` only highlighted `foo` (first text part), not the full search text
- `type:rust` and other key:value constraints weren't stripped
- Tokens starting with `.` were incorrectly treated as constraints
- Backslash-escaped constraints (e.g. `\*.config` as literal search text) weren't handled

This adds `_is_grep_constraint()` that matches the Rust parser's actual GrepConfig rules: extensions (`*.rs`), path segments (`/src/`), exclusions (`!test`), type filters (`type:rust`), and path-oriented globs. Text parts are now joined with space for highlighting, matching `grep_text()` on the Rust side.

Fixes #331

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/crates/fff-nvim/src/lib.rs
+++ b/crates/fff-nvim/src/lib.rs
@@ -5,8 +5,9 @@ use fff::frecency::FrecencyTracker;
 use fff::path_utils::expand_tilde;
 use fff::query_tracker::QueryTracker;
 use fff::{
-    DbHealthChecker, Error, FFFMode, FileSearchConfig, FuzzySearchOptions, PaginationArgs,
-    QueryParser, Score, SearchResult, SharedFrecency, SharedPicker, SharedQueryTracker,
+    DbHealthChecker, Error, FFFMode, FileSearchConfig, FuzzySearchOptions, GrepConfig,
+    PaginationArgs, QueryParser, Score, SearchResult, SharedFrecency, SharedPicker,
+    SharedQueryTracker,
 };
 use mimalloc::MiMalloc;
 use mlua::prelude::*;
@@ -582,6 +583,18 @@ pub fn get_historical_grep_query(_: &Lua, offset: usize) -> LuaResult<Option<Str
         .into_lua_result()
 }
 
+/// Parse a grep query string and return its text portion (with constraints stripped).
+///
+/// Uses the Rust `GrepConfig` parser as the single source of truth, so Lua
+/// code never needs to re-implement constraint detection.
+pub fn parse_grep_query(lua: &Lua, query: String) -> LuaResult<LuaTable> {
+    let parser = QueryParser::new(GrepConfig);
+    let parsed = parser.parse(&query);
+    let table = lua.create_table()?;
+    table.set("grep_text", parsed.grep_text())?;
+    Ok(table)
+}
+
 pub fn wait_for_initial_scan(_: &Lua, timeout_ms: Option<u64>) -> LuaResult<bool> {
     // Extract the scan signal Arc WITHOUT holding the read lock, so the
     // scan thread can acquire the write lock to store its results.
@@ -822,6 +835,7 @@ fn create_exports(lua: &Lua) -> LuaResult<LuaTable> {
     exports.set("health_check", lua.create_function(health_check)?)?;
     exports.set("shorten_path", lua.create_function(shorten_path)?)?;
     exports.set("hex_dump", lua.create_function(hex_dump::hex_dump)?)?;
+    exports.set("parse_grep_query", lua.create_function(parse_grep_query)?)?;
 
     Ok(exports)
 }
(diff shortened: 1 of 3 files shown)


dmtrKovalenko:
I think we should simply expose the parsed query to the lua results to avoid this double reparse which will break once we add a new token type. 

What are you trying to do ?

mvanhorn:
The highlight stripping in the preview was using a regex that didn't account for all the query syntax the Rust grep engine supports (like negation tokens). When the query had special tokens, the preview highlights would be misaligned.

Exposing the parsed query to lua makes sense - that would keep the Rust parser as the single source of truth and avoid any drift. Happy to take that approach instead if you can point me to where the results are serialized for lua.

dmtrKovalenko:
There is lua_types.rs file. But we probably should only do this by request: add a new function that returns a breakdown of a query by string

mvanhorn:
Replaced the Lua-side constraint detection with a new `parse_grep_query()` function in Rust that delegates to `GrepConfig`. The Lua code now calls `fff.parse_grep_query(query).grep_text` to get the search text with constraints stripped.

This removes `_is_grep_constraint()` entirely and keeps the Rust parser as the single source of truth. Also rebased on main.

mvanhorn:
Thanks for the merge, @dmtrKovalenko!

PR #371, opened 2026-04-13
feat: function to re-open the most recent search

justDeeevin:
Replicates the "resume" functionality of telescope and Snacks.nvim.picker (see #159)

Fairly simple to do; just retain the previous state.

### DIFF
--- a/.luarc.json
+++ b/.luarc.json
@@ -4,12 +4,6 @@
     "version": "LuaJIT"
   },
   "workspace": {
-    "library": [
-      "/opt/homebrew/share/nvim/runtime/lua/vim/_meta",
-      "/opt/homebrew/share/nvim/runtime/lua/vim/shared.lua",
-      "${3rd}/luv/library",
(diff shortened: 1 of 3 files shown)


justDeeevin:
Draft because I can't seem to retain the preview... I got rid of line 595, but that isn't cutting it. Unsure what else could be clearing the preview.

I could always opt to update the preview instead but that feels wasteful

dmtrKovalenko:
plz revert the formatting changes

justDeeevin:
Gotcha, wasn't sure whether that would be agreeable so i figured i'd leave it in. Pretty easy to just get rid of the commit b/c jujutsu vcs :3

What do you think of the changes? Any advice or insights about the preview problem?

dmtrKovalenko:
Does it work? Do you still want to proceee with changes

justDeeevin:
Resuming the previous search works for all but the preview, which I'm seeking advice on. It could technically go in as-is but I really want to pin this down.

PR #374, opened 2026-04-14
feat: Add pi fff extension

SamuelLHuber:
Moves https://github.com/SamuelLHuber/pi-fff into this repo.

(note: i've streamlined some of the code so it's ~250 less LOC and tested that locally to verify it works)

what is to verify is npm publishing and installation after distribution.

git publishing is to be verified too. i've added a pi manifest to the root package.json that should allows ```pi install git:github.com/dmtrKovalenko/fff.nvim``` to register the extension so then either pi install npm:@ff-labs/pi-fff or git install should work

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -571,3 +571,16 @@ jobs:
           npm install
           npm run build
           npm publish --tag "$TAG" --access public || echo "Failed to publish @ff-labs/fff-node (may already exist)"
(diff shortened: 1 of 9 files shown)


dmtrKovalenko:
thank you @SamuelLHuber and sorry that it take so much time I a bit updated it but I sent you the the collab invite so feel free to work on it as much as you want

PR #384, opened 2026-04-16
feat: optional vim-mode in prompt input

### LINKED ISSUE #383: Feature: optional vim-mode in prompt input (Esc → normal mode, vim motions)
https://github.com/user-attachments/assets/0cfb73a5-5416-4dc8-8522-5384a93d4b4a

## Motivation

Right now, pressing `<Esc>` in the picker's prompt closes the picker. Users coming from Telescope are used to `<Esc>` dropping the prompt into **normal mode** so they can use vim motions (`j`/`k`/`w`/`b`/`0`/`^`/`gg`/`G`, etc.) on the query and on the result list, and then a second `<Esc>` to close.

This is a quality-of-life feature for vim-heavy users: edit the query with motions, or pause typing and browse results without leaving the picker.

## Proposal

Add an opt-in config flag (default `false` — preserves current behavior):

```lua
vim.g.fff = {
  vim_mode = true,
}
```

When enabled:

- `<Esc>` in insert mode → native `buftype=prompt` switch to normal mode (we stop intercepting it in insert).
- `<Esc>` in normal mode → closes the picker.
- Existing normal-mode bindings (`j`/`k` for list navigation, `focus_list`, `focus_preview`) become reachable.
- Native motions `w`/`b`/`0`/`^`/`h`/`F<x>`/`T<x>`/... work inside the query text.
- A `CursorMoved`/`CursorMovedI` guard clamps the cursor to the first byte after the prompt prefix (`🪿 `), so motions like `0`/`^`/`b` don't land on the prefix.

## Reference

Telescope implements the same flow natively via `buftype=prompt` and a small set of mode-specific mappings — see `lua/telescope/pickers.lua` and `lua/telescope/mappings.lua`. The mechanism here is the same.

## Opt-in / default

Default `false` so existing users see no behavior change. Vim-motion users flip one flag.

fdarian:
Resolves #383.

## Summary

- Adds opt-in `vim_mode` config flag (default `false`) that enables vim-style navigation in the picker prompt, mirroring Telescope's behavior.
- When enabled, `<Esc>` in insert mode falls through to Neovim's native `buftype=prompt` switch to normal mode; a second `<Esc>` closes the picker.
- Existing normal-mode bindings on the input buffer (`j`/`k` list navigation, `focus_list`, `focus_preview`) become reachable. Native motions `w`/`b`/`0`/`^`/`h`/`F<x>`/`T<x>` work on the query text.
- A `CursorMoved`/`CursorMovedI` guard clamps the cursor to the first byte after the prompt prefix (`🪿 `), so `0`/`^`/`b`/etc. can't land on the prefix.

## Changes

- `lua/fff/conf.lua`: add `vim_mode` field to `FffConfig`, default `false`.
- `lua/fff/picker_ui.lua`:
  - Bind `close` to `{i, n}` when `vim_mode = false` (original behavior); `n`-only when `vim_mode = true`.
  - Install the `CursorMoved`/`CursorMovedI` prompt-prefix guard only when `vim_mode = true`.
- `README.md`, `doc/fff.nvim.txt`: document the new option in the complete-options example.

Default-off, so existing users see no behavior change.

## Test plan

- [ ] With `vim_mode = false` (default): `<Esc>` in prompt closes picker (unchanged).
- [ ] With `vim_mode = true`: `<Esc>` in prompt enters normal mode; `j`/`k` scroll the list; `w`/`b` move within query text; `0`/`^` land on first char of query, not on the prefix; second `<Esc>` closes picker; `i`/`a`/`A` return to insert mode.
- [ ] Multi-byte prefix (default `🪿 `) is respected by the cursor guard.

### DIFF
--- a/doc/fff.nvim.txt
+++ b/doc/fff.nvim.txt
@@ -164,6 +164,7 @@ all available options:
         max_results = 100,
         max_threads = 4,
         lazy_sync = true, -- set to false if you want file indexing to start on open
+        vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode (motions like j/k/w/b/0/^), second <Esc> closes the picker
         layout = {
           height = 0.8,
           width = 0.8,
--- a/doc/fff.nvim.txt
+++ b/doc/fff.nvim.txt
@@ -164,7 +164,7 @@ all available options:
         max_results = 100,
         max_threads = 4,
         lazy_sync = true, -- set to false if you want file indexing to start on open
-        vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode (motions like j/k/w/b/0/^), second <Esc> closes the picker
+        prompt_vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode bindings (also allows <leader>p or <leader>l to jump around) the second <Esc> closes the picker
         layout = {
           height = 0.8,
           width = 0.8,
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ require('fff').setup({
     max_results = 100,
     max_threads = 4,
     lazy_sync = true, -- set to false if you want file indexing to start on open
-    vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode (motions like j/k/w/b/0/^), second <Esc> closes the picker
+    prompt_vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode bindings (also allows <leader>p or <leader>l to jump around) the second <Esc> closes the picker
     layout = {
       height = 0.8,
       width = 0.8,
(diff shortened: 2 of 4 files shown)


doc/fff.nvim.txt
```diff
@@ -164,6 +164,7 @@ all available options:
         max_results = 100,
         max_threads = 4,
         lazy_sync = true, -- set to false if you want file indexing to start on open
+        vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode (motions like j/k/w/b/0/^), second <Esc> closes the picker
```
dmtrKovalenko: I mean we are in neovim let's name it promp_vim_mode
```suggestion
        prompt_vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode bindings (also allows <leader>p or <leader>l to jump around) the second <Esc> closes the picker
```

dmtrKovalenko (commented):
I like the idea of a shortcut but I suppose you could do all the same  by remapping the close binding? I just have "jj" mapped to the exti but we'll merge it

dmtrKovalenko (approved):

doc/fff.nvim.txt
fdarian: sure

PR #393, opened 2026-04-19
fix: avoid unicode filepath suffix panic

tmdgusya:
## Summary
- fix a remaining UTF-8 boundary panic in `path_ends_with_suffix()`
- return `false` instead of panicking when a byte-derived suffix offset lands inside a multibyte character
- add regression tests for the helper and for `apply_constraints(Constraint::FilePath(...))`

## Similar PR / duplicate check
I checked existing PRs before opening this:
- closest prior fix: #373 (`fix: Unicode segmentation crash`)
- no open PR currently covers this remaining `Constraint::FilePath` / `path_ends_with_suffix()` panic path

This PR is intentionally narrow: it fixes the unchecked `path[start..]` slice in `crates/fff-core/src/constraints.rs` without changing matching semantics.

## Root cause
`path_ends_with_suffix()` computed:

```rust
let start = path.len() - suffix.len();
```

and then sliced with:

```rust
path[start..]
```

`start` is a byte offset, not guaranteed to be a UTF-8 char boundary. For Unicode filenames, a non-matching suffix can make `start` land inside a multibyte codepoint, which panics before constraint filtering can return `false`.

## Fix
Use `path.get(start..)` instead of unchecked indexing:
- if `start` is not a valid char boundary, return `false`
- otherwise preserve the existing ASCII-insensitive suffix comparison and `/` boundary behavior

## Verification
I avoided using any user-specific filename in tests and instead used synthetic Unicode fixture names.

### Reproduction guard added
New tests in `crates/fff-core/src/constraints.rs`:
- `test_path_ends_with_suffix_does_not_panic_on_unicode_suffix`
- `test_apply_constraints_file_path_with_unicode_suffix`
- `test_path_contains_segment_does_not_panic_on_unicode_segment`

The important regression case uses a synthetic filename such as:
- `data/유니코드_파일_테스트.csv`

and a deliberately non-matching suffix that would previously place the byte offset in the middle of a multibyte character.

### Commands run
```bash
cargo test -p fff-search constraints::tests -- --nocapture
```

### Result
All constraint tests pass locally after the fix, including the new Unicode regression coverage.

## Scope notes
- no parser changes
- no matching behavior expansion
- no Unicode normalization/case-folding changes
- only panic prevention for valid UTF-8 input in the file path suffix constraint path

### DIFF
--- a/crates/fff-core/src/constraints.rs
+++ b/crates/fff-core/src/constraints.rs
@@ -58,16 +58,32 @@ pub fn path_ends_with_suffix(path: &str, suffix: &str) -> bool {
     }
 
     let start = path.len() - suffix.len();
-    let Some(path_suffix) = path.get(start..) else {
-        return false;
-    };
 
-    if !path_suffix.eq_ignore_ascii_case(suffix) {
+    // Must land on a char boundary — multi-byte UTF-8 can make
+    // `path.len() - suffix.len()` land inside a char.
+    if !path.is_char_boundary(start) {
         return false;
     }
 
-    // Exact match, or the character before is /
-    start == 0 || path_bytes[start - 1] == b'/'
+    if !path[start..].eq_ignore_ascii_case(suffix) {
+        return false;
+    }
+
+    // Exact match, or the character before is '/'.
+    // `start` is a char boundary but `start - 1` may be inside a multi-byte
+    // char, so scan backward to find the preceding ASCII byte.
+    if start == 0 {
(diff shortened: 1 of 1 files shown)


dmtrKovalenko (approved):

dmtrKovalenko:
@copilot resolve the merge conflicts in this pull request

dmtrKovalenko (approved):

PR #394, opened 2026-04-19
fix(windows): directorySearch returns 0 results on Windows

schizoidcock:
## Problem

On Windows, `directorySearch` returns 0 results even when directories exist.

## Root Cause Analysis

`Path::to_string_lossy()` returns platform-native separators:
- Unix: `src/components/Button.tsx`
- Windows: `src\components\Button.tsx`

fff uses `rfind('/')` in 10+ locations to extract directory parts:
```rust
let fname_offset = rel_str.rfind('/').map(|i| i + 1).unwrap_or(0);
```

On Windows, `rfind('/')` returns `None` → `fname_offset = 0` → entire path treated as filename → no directory part extracted → `directorySearch` finds nothing.

## Solution

Added `normalize_path_separator()` helper in `path_utils.rs`:

```rust
#[cfg(windows)]
pub fn normalize_path_separator(path: Cow<'_, str>) -> Cow<'static, str> {
    if path.contains('\\') {
        Cow::Owned(path.replace('\\', "/"))
    } else {
        Cow::Owned(path.into_owned())
    }
}

#[cfg(not(windows))]
pub fn normalize_path_separator(path: Cow<'_, str>) -> Cow<'static, str> {
    Cow::Owned(path.into_owned()) // No-op on Unix
}
```

Applied to all 3 path-to-string conversion points:
- `find_file_index` (line 208) - file lookup by path
- `new_with_metadata` (line 311) - runtime file creation  
- `new_from_walk` (line 365) - initial filesystem indexing

## Additional Fix

Removed frecency filter in `score.rs` that was hiding directories without access history:
```rust
// Before: .filter(|d| d.max_access_frecency() > 0)
// After: Include all directories
```

## Steps to Reproduce

1. Build fff on Windows
2. Index a directory with subdirectories
3. Call `directorySearch('component')`
4. Observe 0 results (before fix) vs correct results (after fix)

## Testing

- `cargo build --release` passes
- Manual testing confirms `directorySearch` now returns results on Windows

### DIFF
--- a/crates/fff-core/src/file_picker.rs
+++ b/crates/fff-core/src/file_picker.rs
@@ -205,7 +205,7 @@ impl FileSync {
 
         // Strip base_path prefix to get the relative path.
         let rel_path = match path.strip_prefix(base_path) {
-            Ok(r) => r.to_string_lossy(),
+            Ok(r) => crate::path_utils::normalize_path_separator(r.to_string_lossy()),
             Err(_) => return Err(0),
         };
 
@@ -308,10 +308,9 @@ impl FileItem {
         git_status: Option<Status>,
         metadata: Option<&std::fs::Metadata>,
     ) -> (Self, String) {
-        let relative_path = pathdiff::diff_paths(&path, base_path)
-            .unwrap_or_else(|| path.clone())
-            .to_string_lossy()
-            .into_owned();
+        let path_buf = pathdiff::diff_paths(&path, base_path).unwrap_or_else(|| path.clone());
+        let relative_path =
+            crate::path_utils::normalize_path_separator(path_buf.to_string_lossy()).into_owned();
 
         let (size, modified) = match metadata {
             Some(metadata) => {
@@ -362,7 +361,8 @@ impl FileItem {
         let is_binary = is_known_binary_extension(path);
 
         let rel = pathdiff::diff_paths(path, base_path).unwrap_or_else(|| path.to_path_buf());
-        let rel_str = rel.to_string_lossy().into_owned();
+        let rel_str =
+            crate::path_utils::normalize_path_separator(rel.to_string_lossy()).into_owned();
         let fname_offset = rel_str.rfind('/').map(|i| i + 1).unwrap_or(0) as u16;
 
         let item = Self::new_raw(fname_offset, size, modified, git_status, is_binary);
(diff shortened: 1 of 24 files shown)


crates/fff-core/src/file_picker.rs
```diff
@@ -362,7 +362,7 @@ impl FileItem {
         let is_binary = is_known_binary_extension(path);
 
         let rel = pathdiff::diff_paths(path, base_path).unwrap_or_else(|| path.to_path_buf());
-        let rel_str = rel.to_string_lossy().into_owned();
+        let rel_str = rel.to_string_lossy().replace("\\", "/");  // Normalize for Windows
```
dmtrKovalenko: we should really not do this, instead we have to figure what is underlying reason it can not find anything there is a deepr underlying frecency

crates/fff-core/src/score.rs
dmtrKovalenko, line 401: wooooow this is so bad how did I leave this function

crates/fff-core/src/file_picker.rs
schizoidcock: Im researching a better fix atm

dmtrKovalenko:
I have added some fixes that I think eliminates the need for the normalization could you please pull and check if this is actualy resolves and issue on your end

schizoidcock:
> I have added some fixes that I think eliminates the need for the normalization could you please pull and check if this is actualy resolves and issue on your end

I will check this to test in a moment, but in my end it is working with my fixes already as intended

dmtrKovalenko:
I don't like the normalization of paths. I want to avoid that as much as possible

schizoidcock:
I pulled the latest main and tested on Windows:

```
directorySearch result:
ok: true
items: 0

grep result:
ok: true
items: 55
totalMatched: 55
```

**grep works**, but **directorySearch still returns 0 items** on Windows.

The issue persists without the path normalization fix. The root cause is that `rfind('/')` in Rust expects forward slashes, but `Path::to_string_lossy()` returns backslashes on Windows.

dmtrKovalenko:
Wait but I got rid of all rfind / no?

schizoidcock:
Tested commit 2c837c8 on Windows — **it works!** 🎉

```
directorySearch('tools'):
  ok: true
  items: 3
    1. packages/cli/src/tools/
    2. hooks/
    3. packages/cli/src/utils/

directorySearch('src'):
  ok: true  
  items: 5
```

The rfind fix resolves the issue. I'll remove my path normalization commits since they're no longer needed.

Thanks for the quick fix!

schizoidcock:
Closing — the fix in commit 2c837c8 works. Thanks for resolving this upstream!

dmtrKovalenko:
Wait but we need to make sure it's merged :D

dmtrKovalenko (approved):

PR #397, opened 2026-04-20
fix(parser): accept "git" alias in negated git-status

qu8n:
PR fixes a silent no-op in the query parser: `!git:modified` (and `!git:clean|staged|untracked`, all documented in the README) match every file instead of filtering.

There's a typo in the negation path's key-alias list: `"gi"` instead of `"git"`. This causes `!git:modified` to fall through the key match -> parsed as `Not(Text("git:modified"))` -> no path contains that substring -> nothing filtered.

Also adds `test_negation_git_status_all_key_aliases` covering all four aliases (`status`, `st`, `g`, `git`) under negation. Existing test only covered `status:`.

### DIFF
--- a/crates/fff-query-parser/src/parser.rs
+++ b/crates/fff-query-parser/src/parser.rs
@@ -174,8 +174,8 @@ impl<'a> FFFQuery<'a> {
     /// with the leading `\` stripped, since the backslash is only an escape
     /// signal to the parser and should not appear in the final pattern.
     ///
-    /// `FuzzyQuery::Empty` → empty string  
-    /// `FuzzyQuery::Text("foo")` → `"foo"`  
+    /// `FuzzyQuery::Empty` → empty string
+    /// `FuzzyQuery::Text("foo")` → `"foo"`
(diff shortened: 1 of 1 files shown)


crates/fff-query-parser/src/parser.rs
```diff
@@ -365,7 +365,7 @@ fn parse_token_without_negation<'a, C: ParserConfig>(
                     "type" if config.enable_type_filter() => {
                         return Some(Constraint::FileType(value));
                     }
-                    "status" | "gi" | "g" | "st" if config.enable_git_status() => {
+                    "status" | "st" | "g" | "git" if config.enable_git_status() => {
```
qu8n, line 368: Make this match the positive path at [line 270](https://github.com/dmtrKovalenko/fff.nvim/blob/main/crates/fff-query-parser/src/parser.rs#L270):
```rs
"status" | "st" | "g" | "git" if config.enable_git_status() => {
```

dmtrKovalenko (approved):

Issue #79, opened 2025-08-08
[HIGHEST PRIORITY] Make Ivy layout for dimitar


dmtrKovalenko:
<img width="639" height="292" alt="Image" src="https://github.com/user-attachments/assets/27c00ba4-ea00-4042-b70c-cf3415ffb4f4" />

arnevm123:
This is good enough ;)
```
		opts = {
			prompt = "> ",
			icons = { enabled = false },
			layout = {
				width = 1,
				height = 1,
				prompt_position = "top",
				preview_position = "top",
			},
		},
```

brattonross:
I would unironically like the ability to recreate the ivy style picker layout. Right now I am achieving something like this with the following config, but it feels a little hacky:

```
layout = {
    row = 0.7,
    height = 0.3,
    width = 1,
    prompt_position = "top",
}
```

insipx:
```vim
layout = {
    width = 1.0,
    height = 0.35,
    prompt_position = "top",
    preview_position = "right",
    preview_size = 0.4,
 }
```

i'm using this, all i'm missing is anchoring the box to the bottom of neovim

dmtrKovalenko:
yeah now it's possible and deafult

PR #399, opened 2026-04-20
Add `layout.anchor` option for configurable picker placement for Neovim plugin

n1ghtmare:
Currently the picker always opens centered in the editor. Some users (me... also see: https://github.com/dmtrKovalenko/fff.nvim/issues/79) prefer a drawer-style layout anchored to the bottom left (like mini.pick) or other screen positions. Instead of only adding the option to anchor it to the bottom left like I prefer it, I added options to add it to:

-  `top_left`, `top`, `top_right`, 
- `left`, `center`, `right`
- `bottom_left`, `bottom`, `bottom_right`

The only way to achieve this today is manually calculating `layout.col` and `layout.row` ratios, which is error-prone and doesn't account for the statusline, tabline, or command line height.

Manual `layout.col` and `layout.row` overrides continue to work and take precedence over the anchor defaults.

Here is a screenshot:

image

### DIFF
--- a/lua/fff/picker_ui.lua
+++ b/lua/fff/picker_ui.lua
@@ -115,33 +115,109 @@ local function compute_layout(config)
   local width = math.floor(terminal_width * width_ratio)
   local height = math.floor(terminal_height * height_ratio)
 
-  local col_ratio_default = 0.5 - (width_ratio / 2)
-  local col_ratio = col_ratio_default
+  -- Account for chrome (statusline, tabline, cmdheight) for edge-anchored positions
+  local has_tabline = vim.o.showtabline == 2 or (vim.o.showtabline == 1 and #vim.api.nvim_list_tabpages() > 1)
+  local has_statusline = vim.o.laststatus > 0
+  local top_edge = has_tabline and 1 or 0
+  local bottom_edge = terminal_height - vim.o.cmdheight - (has_statusline and 1 or 0)
+  local usable_height = bottom_edge - top_edge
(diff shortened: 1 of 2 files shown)


lua/fff/picker_ui.lua
```diff
@@ -115,34 +115,110 @@ local function compute_layout(config)
   local width = math.floor(terminal_width * width_ratio)
   local height = math.floor(terminal_height * height_ratio)
 
-  local col_ratio_default = 0.5 - (width_ratio / 2)
-  local col_ratio = col_ratio_default
+  -- Account for chrome (statusline, tabline, cmdheight) for edge-anchored positions
+  local has_tabline = vim.o.showtabline == 2 or (vim.o.showtabline == 1 and #vim.api.nvim_list_tabpages() > 1)
+  local has_statusline = vim.o.laststatus > 0
+  local top_edge = has_tabline and 1 or 0
+  local bottom_edge = terminal_height - vim.o.cmdheight - (has_statusline and 1 or 0)
```
dmtrKovalenko, line 122: isn't there an api to get the full avaialble size? status line can expand e.g. with <C-f> or am I wrong>?

dmtrKovalenko (approved):

lua/fff/picker_ui.lua
n1ghtmare: Honestly, not entirely sure, I kinda copied this "concept" from mini pick (they had something similar).

PR #409, opened 2026-04-21
perf: Reduce amount fs watchers used for detecting criticial git changes

dmtrKovalenko:
@SkeLLLa this should make you happy


SkeLLLa:
Amazing. Thanks for awesome app.

PR #427, opened 2026-04-29
fix(pi-fff): canonicalize find path constraints before query build

edxeth:
## Problem

`pi-fff` can currently false-negative when `find.path` is passed as an absolute in-workspace path.

Repro shape:

```ts
find({
  pattern: "*",
  path: "/tmp/project/.agents/**",
})
```

FFF indexes repo-relative paths like:

```txt
.agents/plans/example.md
```

but `pi-fff` currently forwards the absolute path into the compact FFF query string unchanged:

```txt
/tmp/project/.agents/** *
```

That cannot match the indexed relative paths, so the tool can report no files even when files exist.

There is also a related adapter-level edge case where a simple recursive directory constraint such as:

```txt
.agents/**
```

can false-negative, while the equivalent directory-prefix constraint works:

```txt
.agents/
```

## Fix

Canonicalize `find.path` and `exclude` constraints in `pi-fff` before building the FFF query string:

- absolute paths inside the active workspace become repo-relative constraints
- absolute paths outside the workspace return a clear error
- simple trailing recursive directory globs like `.agents/**` and `src/**` become directory-prefix constraints
- file globs like `src/**/*.ts` are preserved

This keeps the change at the `pi-fff` adapter boundary. It does not change FFF core indexing, ranking, parser behavior, or matching semantics.

## Tests

Added unit coverage for:

- absolute in-workspace path canonicalization
- absolute out-of-workspace rejection
- `.agents/**` -> `.agents/`
- `src/**/*` -> `src/`
- preserving `src/**/*.ts`
- include + exclude query construction

## Validation

Ran:

```bash
bun test packages/pi-fff/test/query.test.ts
bun run --cwd packages/pi-fff typecheck
bun run --cwd packages/fff-node build
bun run --cwd packages/fff-node typecheck
cargo build --release -p fff-c --no-default-features
bun run --cwd packages/fff-node test
bun run --cwd packages/fff-bun typecheck
bun run --cwd packages/fff-bun test
cargo test --workspace --no-default-features
```

Also validated through live pi against the original failure shape:

- stock `pi-fff`: false-negative
- patched `pi-fff`: returns the expected `.agents` files

Note: default `cargo test --workspace` is blocked locally by the Zig/zlob toolchain setup; the no-default workspace suite passes, and this PR only changes TypeScript adapter query construction.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,7 @@ prepare-node: build
 
 test-bun: prepare-bun
 	cd packages/fff-bun && bun test src/
+	cd packages/pi-fff && bun test test/
 
 test-node: prepare-node
 	cd packages/fff-node && npm run build && node test/e2e.mjs
--- a/packages/pi-fff/package.json
+++ b/packages/pi-fff/package.json
@@ -36,6 +36,7 @@
     "access": "public"
   },
   "scripts": {
+    "test": "bun test test/",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -21,6 +21,7 @@ import type {
   SearchResult,
   MixedItem,
 } from "@ff-labs/fff-node";
+import { buildQuery } from "./query";
 
 // ---------------------------------------------------------------------------
 // Constants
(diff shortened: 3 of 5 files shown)


dmtrKovalenko (approved):
very cool thank you

dmtrKovalenko:
this should close 
https://github.com/dmtrKovalenko/fff/pull/427

dmtrKovalenko:
the one little note that the tests would not run in the CI which is a bummer if you can address this it would be perfect

edxeth:
@dmtrKovalenko done, the new tests now run in CI

Issue #432, opened 2026-05-01
pi-fff: `ffgrep` returns false negatives with `path="."`


tekumara:
> *This was generated by AI during triage.*

## Summary

`ffgrep` returns `No matches found` when searching with `path="."`, even when the pattern exists in a file under `app/`.

This is not just surprising: GPT-5.5 will naturally use `path="."` to mean “search the current repo”, then incorrectly conclude that nothing matches.

## Reproduction

Given a repo containing a file under `app/` with the text:

```text
needle
```

This works:

```text
ffgrep(pattern="needle")
```

This also works with a specific directory constraint using a trailing slash:

```text
ffgrep(pattern="needle", path="app/")
```

But these return no results:

```text
ffgrep(pattern="needle", path=".")
ffgrep(pattern="needle", path="app")
```

Actual result:

```text
No matches found
```

## Expected behavior

`path="."` should search the current/project root directory and return matches, equivalent to omitting `path`.

Directory constraints should also work with or without a trailing slash, so `path="app"` should behave like `path="app/"`.

dmtrKovalenko:
I think this was already done what version you are on?

tekumara:
@ff-labs/pi-fff@0.6.4

Issue #425, opened 2026-04-29
regex parse error: unclosed group when searching for strings with special characters in fff.grep


primavistalab:
I found an error when using the fff.grep command. If the search query contains a special character (like an open bracket (, the tool shows a regex parse error. It seems the fff tries to read the query as a regular expression.

```
Called fff.grep({"query":"attachApplicationScoring(","constraints":"src/main/java/com/....","maxResults":20})
  └ ! regex failed: regex parse error:
        attachApplicationScoring(
                                ^
    error: unclosed group, using literal match
```

fff-mcp 0.6.4 (0f5ead1ab5355ea83be1171c54d8d44a65c5f7b4)

codex cli v0.125.0

Issue #423, opened 2026-04-28
pi-fff: absolute paths don't match repo-relative index paths


DaniPopes:
Passing an absolute path to fffind/ffgrep path constraints returns no matches, while the equivalent repo-relative path works. The pi/MCP wrapper should normalize absolute paths under the indexed base path to repo-relative constraints, or reject outside-base paths with a clear error.

I was using pi + gpt 5.5 and noticed that it always tried passing absolute paths but always failing and giving up on using fff entirely.

dmtrKovalenko:
ha that's a good one, it should've been working

Issue #404, opened 2026-04-21
fff not handling sparse checked out repo


leojplin:
When I have a sparse checked out repo, the files in physical disk is less than what git index thinks, fff scans git index and try to update the status for those non-existent-on-disk files, thus producing tons of logs like: `ERROR fff_search::file_picker: Couldn't update the git status for path. `

This makes nvim perceivably slow when opening and using the pickers.

PR #443, opened 2026-05-03
fix: address open issues (308, 311, 400, 403, 404, 423, 425, 432)

dmtrKovalenko:
## Summary

Bundle of low-risk fixes and regression tests for eight open issues:

- **#308** — regression tests for constraint-only git-status queries (`git:modified`, `status:modified`).
- **#311** — accept `pattern` as a serde alias for `query` on `grep` / `find_files`; LLMs that confuse it with `multi_grep`'s `patterns` no longer get a "missing field" error.
- **#400 (issue 3)** — `maxResults: 0` was asymmetric across tools; centralize handling via `normalize_max_results` so 0/negative/non-finite all fall back to the default.
- **#403** — pi-fff `ensureFinder` race: parallel ffgrep/fffind calls could deadlock at the native DB layer. Gate creation on an in-flight promise.
- **#404** — sparse checkouts spammed `ERROR fff_search::file_picker: Couldn't update the git status for path` for every path in the git index that isn't on disk. Demoted to debug.
- **#423** — regression tests for absolute in-workspace path normalization in pi-fff.
- **#425** — stop surfacing `! regex failed: ..., using literal match` alongside successful literal-match results; the error confused LLM callers.
- **#432** — regression tests for `path="."`, `path="./"`, absolute workspace root, and bare directory names in pi-fff.

## Test plan

- [x] `cargo test -p fff-search constraint_only` — 2 passing
- [x] `cargo test -p fff-mcp` — 10 passing
- [x] `cd packages/pi-fff && bun test` — 10 passing
- [x] `cargo clippy --workspace --features zlob -- -D warnings` — clean
- [ ] CI green on this branch


crates/fff-core/src/file_picker.rs
```diff
@@ -1286,7 +1286,10 @@ impl FilePicker {
                         dir.update_frecency_if_larger(score);
                     }
                 } else {
-                    error!(?path, "Couldn't update the git status for path");
+                    // Expected on sparse checkouts: git reports a status for
+                    // a path that isn't materialized on disk and therefore
+                    // isn't in the file index. Don't spam the log (#404).
+                    debug!(?path, "Git status for path not in index, skipping");
```
leojplin, line 1292: does having a sparse checkout repo hinder performance? the intention of my issue is to mitigate the slowness, I imagine we don't need to care about files not present on disk whatsoever. having an else branch here means we still have to do 'something' about it, which is why i think it is slower.

leojplin: if git doesn't return those non checked out files, the code here wouldn't even run right? so i think the root cause is git somehow caring about files not checked out.

dmtrKovalenko: wait are you saying that this issue is not fixed?

leojplin: correct,  i think the root issue has to do with fff doing unnecessary work. when the repo is sparsely checked out, we are  still attempting to update the git status of files that aren't checked out.

PR #444, opened 2026-05-04
feat(mcp): add Windows PowerShell installer

### LINKED ISSUE #364: Windows Support
### Summary

Currently from the install for AI agents it seems like it is only support for UNIX based devices based on the install script. However, on the releases it seems like there is Windows Support?

Please impl, a install (most likely PowerShell) script so Windows users can install this package / thing :)

mynameistito:
## Summary

Closes #364.

Adds `install-mcp.ps1`, a PowerShell mirror of `install-mcp.sh`, so Windows users can install the `fff-mcp` binary with a one-liner. README updated with the Windows command alongside the existing Unix one.

The release workflow already publishes `fff-mcp-x86_64-pc-windows-msvc.exe` and `fff-mcp-aarch64-pc-windows-msvc.exe`, so no CI changes needed — the script just consumes existing assets.

### What it does
- Detects AMD64 / ARM64 (honoring `PROCESSOR_ARCHITEW6432` for WOW64).
- Picks the latest release containing the matching asset (mirrors the bash awk logic via `Invoke-RestMethod`).
- Downloads + verifies SHA256 (skips with a warning if `.sha256` is missing, same as bash).
- Installs to `$env:LOCALAPPDATA\fff-mcp\bin` (overridable via `FFF_MCP_INSTALL_DIR`).
- Adds install dir to user `PATH` if not already present.
- Detects `claude` / `opencode` / `codex` and prints the same wiring snippets as the bash script.
- Update path: re-runs in-place if binary already exists, prints succinct \"updated to \<tag\>\" message.

### Invocation

```powershell
irm https://raw.githubusercontent.com/dmtrKovalenko/fff.nvim/main/install-mcp.ps1 | iex
```

(Eventually you may want to host this on `dmtrkovalenko.dev/install-fff-mcp.ps1` for symmetry — README points at the raw GitHub URL for now.)

## Checklist

- [x] No changes to Rust / Lua / public APIs
- [x] No CI workflow changes (existing Windows release assets reused)
- [x] PowerShell 5.1+ compatible (ships with Windows 10/11 — no extra deps)
- [x] Honors `FFF_MCP_INSTALL_DIR` env override (parity with bash)
- [x] SHA256 verification when checksum file present
- [x] Idempotent: re-running upgrades in place
- [x] Adds install dir to user PATH only if missing
- [x] README updated with both Linux/macOS and Windows install snippets
- [x] Script syntax validated via `[System.Management.Automation.PSParser]::Tokenize`
- [ ] End-to-end install tested on a fresh Windows machine (deferred — needs a real release with Windows assets to download)
- [ ] ARM64 Windows tested (no hardware available)

## Test plan

- [x] On Windows AMD64: run the one-liner, confirm `fff-mcp.exe` lands in `%LOCALAPPDATA%\fff-mcp\bin`, confirm it is on PATH in a new shell, confirm `fff-mcp --help` (or equivalent) runs.
- [x] Re-run installer, confirm \"Updating...\" path triggers and binary is replaced.
- [x] Set `$env:FFF_MCP_INSTALL_DIR` to a custom dir, confirm install lands there.
- [x] Delete the `.sha256` upstream temporarily (or test against a release missing it) and confirm the script warns instead of failing.
- [x] Confirm Claude Code / OpenCode / Codex detection prints the right snippet when those CLIs are on PATH.

### DIFF
--- a/install-mcp.ps1
+++ b/install-mcp.ps1
@@ -13,11 +13,19 @@
     Release tag to install (e.g. 'v0.1.2'). Default: latest release containing a Windows MCP asset.
 .PARAMETER InstallDir
     Target install directory. Default: $env:LOCALAPPDATA\fff-mcp\bin.
+.PARAMETER PathScope
+    How to persist PATH: 'User' (set user env var, default), 'Profile' (append to $PROFILE *nix-style), 'None' (do not persist).
+    Env-var fallback: $env:FFF_MCP_PATH_SCOPE.
 #>
 param(
     [string]$Version = $env:FFF_MCP_VERSION,
-    [string]$InstallDir = $env:FFF_MCP_INSTALL_DIR
+    [string]$InstallDir = $env:FFF_MCP_INSTALL_DIR,
+    [ValidateSet('User', 'Profile', 'None')]
+    [string]$PathScope
 )
+if (-not $PathScope) {
+    $PathScope = if ($env:FFF_MCP_PATH_SCOPE) { $env:FFF_MCP_PATH_SCOPE } else { 'User' }
+}
 
 $ErrorActionPreference = 'Stop'
 
@@ -124,7 +132,30 @@ function Add-ToUserPath {
         [Environment]::SetEnvironmentVariable('Path', $newPath, 'User')
         Write-Success "Added $Dir to user PATH."
     }
-    # Make available in current session too — no shell restart needed.
+}
+
+function Add-ToProfilePath {
+    param([string]$Dir)
+    $profilePath = $PROFILE.CurrentUserAllHosts
+    $line = "`$env:PATH += ';$Dir'  # added by fff-mcp installer"
+    if (Test-Path $profilePath) {
+        $existing = Get-Content $profilePath -Raw -ErrorAction SilentlyContinue
+        if ($existing -and $existing.Contains($Dir)) { return }
+    } else {
+        New-Item -ItemType File -Force -Path $profilePath | Out-Null
+    }
+    Add-Content -Path $profilePath -Value "`n$line"
+    Write-Success "Appended PATH update to $profilePath."
+}
+
+function Set-Path {
+    param([string]$Dir, [string]$Scope)
+    switch ($Scope) {
+        'User'    { Add-ToUserPath $Dir }
+        'Profile' { Add-ToProfilePath $Dir }
+        'None'    { Write-Info "Skipping PATH persistence (-PathScope None)." }
+    }
+    # Make available in current session regardless of scope.
     if (-not (Test-OnPath $Dir)) { $env:PATH = "$env:PATH;$Dir" }
 }
 
@@ -225,7 +256,7 @@ function Main {
         Write-Success "FFF MCP Server updated to $tag!"
         Write-Host ""
     } else {
-        Add-ToUserPath $InstallDir
+        Set-Path -Dir $InstallDir -Scope $PathScope
         Show-SetupInstructions -BinaryPath $binaryPath
     }
 }
--- a/install-mcp.ps1
+++ b/install-mcp.ps1
@@ -8,6 +8,9 @@
 
 $ErrorActionPreference = 'Stop'
 
+# Force TLS 1.2 — PS 5.1 on older Win10 may default to SSL3/TLS1.0 which GitHub rejects.
+[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
 $Repo = 'dmtrKovalenko/fff.nvim'
 $BinaryName = 'fff-mcp'
 $InstallDir = if ($env:FFF_MCP_INSTALL_DIR) { $env:FFF_MCP_INSTALL_DIR } else { Join-Path $env:LOCALAPPDATA 'fff-mcp\bin' }
@@ -45,7 +48,6 @@ function Install-Binary {
 
     $filename = "$BinaryName-$Target.exe"
     $url = "https://github.com/$Repo/releases/download/$Tag/$filename"
-    $checksumUrl = "$url.sha256"
 
     Write-Info "Downloading $filename from release $Tag..."
 
@@ -63,19 +65,6 @@ function Install-Binary {
             Write-Host "  Platform: $Target"
             Write-Host "Check available releases at: https://github.com/$Repo/releases"
             throw
-        }
-
-        $tmpSha = "$tmpFile.sha256"
-        try {
-            Invoke-WebRequest -Uri $checksumUrl -OutFile $tmpSha -UseBasicParsing -ErrorAction Stop
-            Write-Info "Verifying checksum..."
-            $expected = (Get-Content $tmpSha -Raw).Trim().Split()[0]
-            $actual = (Get-FileHash -Path $tmpFile -Algorithm SHA256).Hash.ToLower()
-            if ($expected.ToLower() -ne $actual) {
-                throw "Checksum verification failed! expected=$expected actual=$actual"
-            }
-        } catch [System.Net.WebException] {
-            Write-Warn "Checksum file not available, skipping verification."
         }
 
         New-Item -ItemType Directory -Force -Path $InstallDir | Out-Null
--- a/install-mcp.ps1
+++ b/install-mcp.ps1
@@ -3,8 +3,21 @@
 .SYNOPSIS
     FFF MCP Server installer for Windows.
 .DESCRIPTION
-    Usage: irm https://raw.githubusercontent.com/dmtrKovalenko/fff.nvim/main/install-mcp.ps1 | iex
+    Pipe usage:
+        irm https://raw.githubusercontent.com/dmtrKovalenko/fff.nvim/main/install-mcp.ps1 | iex
+    Direct usage (supports params):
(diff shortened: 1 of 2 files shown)


mynameistito:
All done with Claude Opus 4.7 and worked as intended for me :)

Did not read any of the code, just tested the outcome happy to make itterations should you wish @dmtrKovalenko

dmtrKovalenko:
I have no idea is this right or not and do not understand anything in powershell but my opus 4.7 told this LMAO

```
  - TLS 1.2 not forced. install-mcp.ps1:33 calls Invoke-RestMethod against api.github.com without setting      
  [Net.ServicePointManager]::SecurityProtocol. On PS 5.1 (the minimum version required) on older Win10 builds,
  SecurityProtocol can still default to SSL3/TLS1.0, and GitHub rejects those. Common pitfall in PowerShell    
  installers — recommend adding near the top:
  [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor
  [Net.SecurityProtocolType]::Tls12
  - Wrong exception type for missing SHA256. install-mcp.ps1:66 catches [System.Net.WebException], but on
  PowerShell 7+ a 404 from Invoke-WebRequest throws Microsoft.PowerShell.Commands.HttpResponseException, not
  WebException. Since #Requires -Version 5.1 admits PS 7, and $ErrorActionPreference = 'Stop', a missing
  checksum file will abort the install instead of warning — the exact opposite of the bash script's behavior.
  Use a generic catch or catch both types:
  } catch {
      if ($_.Exception -is [System.Net.WebException] -or $_.Exception.GetType().Name -eq
  'HttpResponseException') {
          Write-Warn "Checksum file not available, skipping verification."
      } else { throw }
  }
```

I think we should get rid of sha256 check at all

mynameistito:
that was what my opus thought, maybe good to get other eyes on it but it works as intended for me :)

mynameistito:
Both points valid — pushed db25a30:

- Forced TLS 1.2 at top of script via `[Net.ServicePointManager]::SecurityProtocol = ... -bor [Net.SecurityProtocolType]::Tls12`.
- Dropped sha256 verification entirely per your call. Kills the WebException vs HttpResponseException mismatch and halves the script's failure surface. Bash version's check was best-effort anyway.

install-mcp.ps1
```diff
@@ -0,0 +1,196 @@
+#Requires -Version 5.1
+<#
+.SYNOPSIS
+    FFF MCP Server installer for Windows.
+.DESCRIPTION
+    Usage: irm https://raw.githubusercontent.com/dmtrKovalenko/fff.nvim/main/install-mcp.ps1 | iex
+#>
```
ccastanedaucf, line 19: could be nice to have some script-level args so it's easy to rollback or install to a custom dir, like
```pwsh
.PARAMETER Version
    blah blah blah
.PARAMETER InstallDir
    blah blah blah
#>
param(
    [string]$Version = 'latest',
    [string]$InstallDir
)
```

install-mcp.ps1
```diff
@@ -0,0 +1,196 @@
...
+$BinaryName = 'fff-mcp'
+$InstallDir = if ($env:FFF_MCP_INSTALL_DIR) { $env:FFF_MCP_INSTALL_DIR } else { Join-Path $env:LOCALAPPDATA 'fff-mcp\bin' }
+
+function Write-Info    { param($m) Write-Host $m -ForegroundColor Blue }
+function Write-Success { param($m) Write-Host $m -ForegroundColor DarkYellow }
+function Write-Warn    { param($m) Write-Host $m -ForegroundColor Yellow }
+
+function Get-Target {
+    $arch = $env:PROCESSOR_ARCHITECTURE
+    if ($env:PROCESSOR_ARCHITEW6432) { $arch = $env:PROCESSOR_ARCHITEW6432 }
```
ccastanedaucf: ya there's some weirdness with env vars due to ARM emulation, hitting the registry or CMI/WMI is the most reliable (this is what bun does for example)
```pwsh
$arch = (Get-ItemProperty 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment').PROCESSOR_ARCHITECTURE
```

install-mcp.ps1
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+function Get-LatestReleaseTag {
+    param([string]$Target)
+    $asset = "$BinaryName-$Target.exe"
+    $headers = @{ 'User-Agent' = 'fff-mcp-installer' }
+    if ($env:GITHUB_TOKEN) { $headers['Authorization'] = "Bearer $env:GITHUB_TOKEN" }
+
+    $releases = Invoke-RestMethod -Uri "https://api.github.com/repos/$Repo/releases" -Headers $headers
```
ccastanedaucf, line 59: maybe just hit `/latest/` or have a script-level version parameter so this doesn't have to pull down all releases

install-mcp.ps1
```diff
@@ -0,0 +1,196 @@
...
+    $url = "https://github.com/$Repo/releases/download/$Tag/$filename"
+
+    Write-Info "Downloading $filename from release $Tag..."
+
+    $tmp = Join-Path ([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
+    New-Item -ItemType Directory -Force -Path $tmp | Out-Null
+    try {
+        $tmpFile = Join-Path $tmp $filename
+        try {
+            Invoke-WebRequest -Uri $url -OutFile $tmpFile -UseBasicParsing
```
ccastanedaucf: curl ships with windows since 2018 and will be faster than 5.1's iwr, but should still have a fallback like
```pwsh
if (Get-Command curl.exe -ErrorAction SilentlyContinue) {
    curl.exe -#SfLo $tmpFile $url
else {
# ... Invoke-WebRequest path
}
```

Also IIRC progress bars have some awful perf issues on 5.1 that slows download speed, so most scripts will do something like
```pwsh
$prevProgressPref = $ProgressPreference
try {
    $ProgressPreference = 'SilentlyContinue'
    # do stuff
} finally {
    $ProgressPreference = $prevProgressPref
}

File: install-mcp.ps1 [hunk 5]
```diff
@@ -0,0 +1,196 @@
...
+
+function Add-ToUserPath {
+    param([string]$Dir)
+    $userPath = [Environment]::GetEnvironmentVariable('Path', 'User')
+    if (-not $userPath) { $userPath = '' }
+    $entries = $userPath -split ';' | Where-Object { $_ -ne '' }
+    if ($entries -notcontains $Dir) {
+        $newPath = (@($entries + $Dir) -join ';')
+        [Environment]::SetEnvironmentVariable('Path', $newPath, 'User')
+        Write-Success "Added $Dir to user PATH (open a new shell to pick it up)."
```
ccastanedaucf: you can make this available on the current session's path too
```pwsh
$env:PATH += ";$Dir"
```

ccastanedaucf (commented):
lgtm

mynameistito:
Thanks @ccastanedaucf — pushed 47d6006 with all five:

- **Params**: added `-Version` and `-InstallDir`. Pipe form keeps working via `FFF_MCP_VERSION` / `FFF_MCP_INSTALL_DIR` env-var fallbacks, since `irm | iex` can't bind params.
- **Arch detect**: switched to `HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment`, dropped the env-var dance.
- **Release scan**: kept the asset-aware scan as the default path (the bash script does the same — not every release publishes the MCP exe), but `-Version` short-circuits it.
- **Download**: `curl.exe` primary with iwr fallback; iwr path wraps `$ProgressPreference = 'SilentlyContinue'` in try/finally to keep PS 5.1 throughput sane.
- **PATH**: also appends to `$env:PATH` for the current session so no shell restart is needed.

install-mcp.ps1
```diff
@@ -0,0 +1,233 @@
...
+    if (-not $userPath) { $userPath = '' }
+    $entries = $userPath -split ';' | Where-Object { $_ -ne '' }
+    if ($entries -notcontains $Dir) {
+        $newPath = (@($entries + $Dir) -join ';')
+        [Environment]::SetEnvironmentVariable('Path', $newPath, 'User')
+        Write-Success "Added $Dir to user PATH."
+    }
+    # Make available in current session too — no shell restart needed.
+    if (-not (Test-OnPath $Dir)) { $env:PATH = "$env:PATH;$Dir" }
+}
```
PurpleMyst, line 160: For what it's worth, some pwsh users such as myself prefer appending to $profile *nix-style rather than modifying the environment variable globally; there could probably be a script param to control this behavior.

mynameistito: thanks for the feedback pushing another commit out now.

mynameistito:
Thanks @PurpleMyst — pushed 61fc7e8 with a `-PathScope` param:

- `-PathScope User` (default) — sets the user env var, current behavior.
- `-PathScope Profile` — appends `$env:PATH += ';...'` to `$PROFILE.CurrentUserAllHosts`, *nix-style.
- `-PathScope None` — skip persistence entirely.

Env-var fallback `FFF_MCP_PATH_SCOPE` for the `irm | iex` case. Current-session `$env:PATH` is still updated regardless of scope so the binary is usable immediately.

PurpleMyst:
LGTM

PR #454, opened 2026-05-06
fix(pi-fff): avoid editor overwrite when tools-only mode is enabled

mrjoedang:
## Summary
- `pi-fff` currently overwrites editor/statusline components installed by other extensions after startup even when `tools-only` mode is enabled
- this skips editor changes on startup when mentions are disabled (`tools-only`)


dmtrKovalenko (approved):

SamuelLHuber:
thank you, tested your branch and works as expected for me locally.

PR #461, opened 2026-05-09
fix: the appearance of indents when height = 1.0 or width = 1.0

### LINKED ISSUE #366: When the height is equal to 1, the first row of the screen is not used.
screenshots with different positions for the prompt and preview:

Image

Image

Image

Image

minimal neovim config:
```lua
vim.pack.add { 'https://github.com/dmtrKovalenko/fff.nvim' }

vim.api.nvim_create_autocmd('PackChanged', {
    callback = function(ev)
        local name, kind = ev.data.spec.name, ev.data.kind
        if name == 'fff.nvim' and (kind == 'install' or kind == 'update') then
            if not ev.data.active then
                vim.cmd.packadd 'fff.nvim'
            end
            require('fff.download').download_or_build_binary()
        end
    end,
})

require('fff').setup {
    layout = {
        height = 1,
        width = 1,
        prompt_position = 'bottom',
        preview_position = 'left',
    },
}

vim.g.mapleader = ' '

vim.keymap.set('n', '<leader>ff', function()
    require('fff').find_files()
end, {
    silent = true,
    desc = 'FFF Find Files',
})

vim.keymap.set('n', '<leader>lg', function()
    require('fff').live_grep()
end, {
    silent = true,
    desc = 'Snacks Picker Grep',
})
```

astrelinvl:
fix https://github.com/dmtrKovalenko/fff/issues/366

### DIFF
--- a/lua/fff/picker_ui.lua
+++ b/lua/fff/picker_ui.lua
@@ -170,8 +170,8 @@ local function compute_layout(config)
   --   col: -1 for left (internal +1 on list_col makes it flush)
   --        -2 for right (internal +1 plus the preview window's independent right border)
   --   row: -1 for top/bottom (internal +1 on rows; bottom also accounts for chrome via bottom_edge)
-  local center_col = math.floor((terminal_width - width) / 2)
-  local center_row = top_edge + math.floor((usable_height - height) / 2)
+  local center_col = math.floor((terminal_width - width) / 2) - 1
+  local center_row = top_edge + math.floor((usable_height - height) / 2) - 1
   local anchor_positions = {
     center = {
       col = center_col,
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
The problem with your change is that they apply to all sizes, not just fullscreen. Consider  
            width = 0.8, height = 0.8 (the defaults):  
                - center_col would be calculated as math.floor((200 - 160) / 2) - 1 = 19 instead of 20 — the picker shifts one cell left  
                - center_row shifts one cell upward similarly  
                - total_height gets an extra row that wasn't taken into account in the height calculation

dmtrKovalenko:
can you please check on the main if you still have this issue?

astrelinvl:
When 
prompt_position = 'top',
preview_position = 'right',

or


prompt_position = 'top',
preview_position = 'bottom',

or

prompt_position = 'bottom',
preview_position = 'bottom',

statusline is visible + borders are drawn incorrectly

PR #462, opened 2026-05-09
Feat: Add an option to make changing window directory with `lcd` work.

mkblast:
I think the best way to explain this PR is to show my workflow.

I have this `autocmd` where I `lcd` into the directory I navigate to using `oil.nvim`:
```lua
local oil_local_cwd = vim.api.nvim_create_augroup("OilLocalCwd", { clear = true })
autocmd("BufEnter", {
    group = oil_local_cwd,
    callback = function(o)
        if o.match:find("^oil://") then
            vim.cmd.lcd(require("oil").get_current_dir())
        else
            vim.cmd.lcd(vim.fn.getcwd(-1))
        end
    end,
    nested = true
})
```
This is mainly for when I execute a command it execute in that directory. However, this breaks `fff` when I search for a file from `oil.nvim`:

image
 
With this patch I introduced a new config option `change_local_directory` (I couldn't find a better name) that uses the window directory.

I hope that was clear.

### DIFF
--- a/lua/fff/conf.lua
+++ b/lua/fff/conf.lua
@@ -199,7 +199,7 @@ local function init()
     max_threads = 4,
     lazy_sync = true, -- set to false if you want file indexing to start on open
     prompt_vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode bindings (also allows <leader>p or <leader>l to jump around) the second <Esc> closes the picker
-    change_index_cwd_on_lcd = false,
+    change_index_cwd_on_lcd = false, -- set to true if you want the picker to index from the current local working directory (:h lcd).
     layout = {
       height = 0.8,
       width = 0.8,
--- a/lua/fff/conf.lua
+++ b/lua/fff/conf.lua
@@ -197,9 +197,9 @@ local function init()
     title = 'FFFiles',
     max_results = 100,
     max_threads = 4,
-    change_local_directory = false,
     lazy_sync = true, -- set to false if you want file indexing to start on open
     prompt_vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode bindings (also allows <leader>p or <leader>l to jump around) the second <Esc> closes the picker
+    change_index_cwd_on_lcd = false,
     layout = {
       height = 0.8,
       width = 0.8,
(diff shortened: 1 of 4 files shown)


lua/fff/core.lua
```diff
@@ -46,10 +46,7 @@ local function setup_global_autocmds(config)
   vim.api.nvim_create_autocmd('DirChanged', {
     group = group,
     callback = function()
-      -- Window-local `:lcd` / `:tcd` are per-window — they don't change the
-      -- effective project root for the picker, so bail before touching
-      -- anything else.
-      if vim.v.event.scope == 'window' then return end
+      if vim.v.event.scope == 'window' and config.change_local_directory == false then return end
```
dmtrKovalenko: not sure if the option name actually reflects going? you want the lcd to change the picker as well why?

The other windows will also change the picker becuase there is only one ongoing instance available.

If you still insist on this change let's do something like `change_index_cwd_on_lcd: true` api which is more clear what's going on

dmtrKovalenko (changes requested):

lua/fff/core.lua
mkblast: Thanks for the reply! :D

I'll describe my use case to see if it fits the project. I mainly use this when I'm browsing a library or the `std` (primarily `C/C++ headers`) without a language server.

I usually have one window where its `lcd` is the root project, and another window where its `lcd` is a third party library path for greping/searching symbols or files. This workflow makes it easy to swap back and forth and has surprisingly removed my need for a language server.

> If you still insist on this change let's do something like change_index_cwd_on_lcd: true api which is more clear what's going on.

I’m aware the naming was a bit wack. I was just trying to see if the patch would even be considered! I’ll be happy to change the option name to something clearer.

lua/fff/conf.lua
```diff
@@ -199,7 +199,7 @@ local function init()
     max_threads = 4,
     lazy_sync = true, -- set to false if you want file indexing to start on open
     prompt_vim_mode = false, -- set to true to enable vim-mode in the prompt: <Esc> leaves insert for normal mode bindings (also allows <leader>p or <leader>l to jump around) the second <Esc> closes the picker
-    change_index_cwd_on_lcd = false,
+    change_index_cwd_on_lcd = false, -- set to true if you want the picker to index from the current local working directory (:h lcd).
```
dmtrKovalenko, line 202: I think you lost all your other changes 😭

dmtrKovalenko (approved):

mkblast:
Yeah... I wasn't able to figure it out why it doesn't work...
I'll close this pr for now. I might take another shot if I find more time later

PR #464, opened 2026-05-10
feat: add Python SDK (pip install fff-search)

YounesFil:
The README mentions Python via ctypes for binding the C lib but there's no Python package. Made one.

`crates/fff-py` is a small PyO3 crate that wraps `fff-search` directly. Installs as `pip install fff-search`. The API mirrors the Node SDK so it should look familiar:

```python
from fff_search import FileFinder

with FileFinder.create(base_path=".", ai_mode=True) as f:
    f.wait_for_scan(timeout_ms=10_000)
    files = f.file_search("incognito profile", page_size=20)
    hits = f.grep("GetOffTheRecordProfile", classify_definitions=True)
```

In the PR:
- `crates/fff-py/`: the new crate
- `python/fff_search/`: Python facade with type stubs
- `tests/`: 18 pytest tests covering search, grep, lifecycle, pagination
- `.github/workflows/python-wheels.yml`: builds wheels for linux/musllinux/macos/windows on tag, smoke tests them, publishes to PyPI

Using `abi3-py39` so one wheel per platform handles Python 3.9 through 3.13.

Tested locally on Windows. `cargo check` is clean, `maturin develop` builds, all 18 pytest tests pass.

Couple things before merge:
1. Went with PyO3 because it gives real typed Python objects (FileItem, GrepMatch etc) instead of forcing users to deal with raw ctypes. If you'd rather match fff-node and use ctypes over libfff_c, I can rewrite it.
2. Named the package fff-search to match the Rust crate. If you want something different, or want me to grab fff on PyPI too, lmk.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -1,0 +1,166 @@
+name: python-wheels
+
+on:
+  push:
+    tags:
+      - "fff-py-v*"
+  workflow_dispatch:
+    inputs:
+      publish:
+        description: "Publish to PyPI (requires PYPI_API_TOKEN secret)"
+        required: false
+        default: "false"
+
+permissions:
+  contents: read
+
+jobs:
+  linux:
+    name: linux ${{ matrix.target }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [x86_64, aarch64]
+    steps:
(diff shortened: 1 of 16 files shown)


dmtrKovalenko:
why did you close the PR? I think it's cool

YounesFil:
closed it to fix a couple things, will reopen soon. been using fff for a while now, really like it.

PR #478, opened 2026-05-18
fix(pi-fff): avoid blocking session start during finder warmup

### LINKED ISSUE #477: pi-fff blocks Pi /new and /resume while waiting for FileFinder scan on session_start
## Summary

`@ff-labs/pi-fff` currently blocks Pi session replacement (`/new` and `/resume`) because the `session_start` handler eagerly initializes `FileFinder` and awaits the initial scan.

In `packages/pi-fff/src/index.ts`:

```ts
pi.on("session_start", async (_event, ctx) => {
  try {
    activeCwd = ctx.cwd;
    if (shouldEnableMentions()) applyEditorMode(ctx);
  } catch (e: unknown) {
    ctx.ui.notify(...);
  }
});
```

`ensureFinder()` calls:

```ts
const result = FileFinder.create({ basePath: cwd, ... });
...
await finder.waitForScan(15000);
```

Because Pi recreates/reloads the session runtime when switching sessions, this means `/new` and selecting a session from `/resume` can hang until `FileFinder.create()` / `waitForScan(15000)` finishes.

## Impact

- `/resume`: after selecting a session, Pi appears stuck before the resumed session is usable.
- `/new`: after pressing Enter, Pi appears stuck before the new session is usable.
- This is especially noticeable in larger workspaces where scanning/watching takes non-trivial time.
- Disabling `@ff-labs/pi-fff` makes `/new` and `/resume` fast again.

## Expected behavior

Pi session startup should not be blocked by a file index warmup. The finder can still be warmed in the background, and actual fff tool calls / @-mention autocomplete can await the same in-flight initialization if needed.

## Suggested fix

Make the `session_start` warmup non-blocking, e.g. schedule it after the session start event returns:

```ts
let lifecycleId = 0;

pi.on("session_start", async (_event, ctx) => {
  try {
    activeCwd = ctx.cwd;
    const sessionLifecycleId = ++lifecycleId;
    if (shouldEnableMentions()) applyEditorMode(ctx);

    setTimeout(() => {
      if (sessionLifecycleId !== lifecycleId) return;
      ensureFinder(activeCwd).catch((e: unknown) => {
        if (sessionLifecycleId !== lifecycleId) return;
        ctx.ui.notify(
          `FFF init failed: ${e instanceof Error ? e.message : String(e)}`,
          "error",
        );
      });
    }, 0);
  } catch (e: unknown) {
    ctx.ui.notify(
      `FFF init failed: ${e instanceof Error ? e.message : String(e)}`,
      "error",
    );
  }
});

pi.on("session_shutdown", async () => {
  lifecycleId++;
  destroyFinder();
});
```

I also found it safer for `ensureFinder()` to return the local `createdFinder` after `waitForScan()` rather than the mutable global `finder`, so a shutdown during warmup cannot accidentally return a destroyed/replaced finder.

## Local validation

I tested a local patch with:

```bash
cd packages/pi-fff
npx tsc --noEmit
```

and verified that importing the extension still returns a valid factory. With the patch, `/new` and `/resume` are no longer blocked by the scan warmup; the first actual fff operation may still await initialization, which seems like the expected tradeoff.

## Related issues I found

I did not find an exact duplicate for `pi-fff` blocking Pi `/new` or `/resume`. Some related `pi-fff`/watcher issues:

- #403: parallel `ffgrep` calls and `ensureFinder` singleton synchronization
- #439: `aiMode: true` watcher file descriptor usage on macOS
- #476: macOS segfault while watching directories during clone operations

This issue seems specifically about lifecycle behavior in the Pi extension: doing blocking index warmup inside `session_start`.

## Environment

- `@ff-labs/pi-fff`: 0.8.1
- Pi: 0.75.1
- Node.js: v24.13.1
- OS: macOS / Darwin arm64

cokekitten:
## Summary

Fixes #477.

`pi-fff` currently awaits `ensureFinder()` during `session_start`, and `ensureFinder()` calls `FileFinder.create()` followed by `waitForScan(15000)`. Since Pi recreates the session runtime for `/new` and `/resume`, this can make session switching appear stuck while the finder scan warms up.

This change:

- schedules finder warmup after `session_start` returns, so Pi session replacement is not blocked
- keeps tool behavior intact: the first fff tool/autocomplete use still awaits the shared in-flight `finderPromise` if warmup is not done yet
- adds a lifecycle guard so stale warmups from a previous session do not notify or affect a newer session
- returns the local `createdFinder` from `ensureFinder()` after `waitForScan()` to avoid returning a destroyed/replaced global finder if shutdown happens during warmup

## Validation

```bash
cd packages/pi-fff
npx tsc --noEmit
```

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -292,6 +292,7 @@ export default function fffExtension(pi: ExtensionAPI) {
   // deadlock at the native layer (issue #403).
   let finderPromise: Promise<FileFinder> | null = null;
   let activeCwd = process.cwd();
+  let lifecycleId = 0;
 
   // Mode resolution: flag > env > default
   let currentMode: FffMode =
@@ -345,10 +346,14 @@ export default function fffExtension(pi: ExtensionAPI) {
       if (!result.ok)
         throw new Error(`Failed to create FFF file finder: ${result.error}`);
 
-      finder = result.value;
+      const createdFinder = result.value;
+      finder = createdFinder;
       finderCwd = cwd;
-      await finder.waitForScan(15000);
-      return finder;
+      await createdFinder.waitForScan(15000);
+      if (createdFinder.isDestroyed || finder !== createdFinder || finderCwd !== cwd) {
+        throw new Error("FFF finder was destroyed before initialization completed");
+      }
+      return createdFinder;
     })().finally(() => {
       finderPromise = null;
     });
@@ -479,8 +484,21 @@ export default function fffExtension(pi: ExtensionAPI) {
   pi.on("session_start", async (_event, ctx) => {
     try {
       activeCwd = ctx.cwd;
+      const sessionLifecycleId = ++lifecycleId;
       if (shouldEnableMentions()) applyEditorMode(ctx);
-      await ensureFinder(activeCwd);
+      // Warm the finder after session startup returns. Tool calls still await
+      // the same in-flight promise on first use, but /new and /resume are not
+      // blocked by FileFinder.create() or waitForScan().
+      setTimeout(() => {
+        if (sessionLifecycleId !== lifecycleId) return;
+        ensureFinder(activeCwd).catch((e: unknown) => {
+          if (sessionLifecycleId !== lifecycleId) return;
+          ctx.ui.notify(
+            `FFF init failed: ${e instanceof Error ? e.message : String(e)}`,
+            "error",
+          );
+        });
+      }, 0);
     } catch (e: unknown) {
       ctx.ui.notify(
         `FFF init failed: ${e instanceof Error ? e.message : String(e)}`,
@@ -490,6 +508,7 @@ export default function fffExtension(pi: ExtensionAPI) {
   });
 
   pi.on("session_shutdown", async () => {
+    lifecycleId++;
     destroyFinder();
   });
 


packages/pi-fff/src/index.ts
```diff
@@ -345,10 +346,14 @@ export default function fffExtension(pi: ExtensionAPI) {
       if (!result.ok)
         throw new Error(`Failed to create FFF file finder: ${result.error}`);
 
-      finder = result.value;
+      const createdFinder = result.value;
+      finder = createdFinder;
       finderCwd = cwd;
-      await finder.waitForScan(15000);
-      return finder;
+      await createdFinder.waitForScan(15000);
```
dmtrKovalenko, line 352: to actually get rid of the wait we proably should redesign the waitForScan internally, it is anyway blocking now while it should do the pool for is_scanning

dmtrKovalenko (approved):

packages/pi-fff/src/index.ts
cokekitten: Agreed. This PR only removes the `session_start` blocking path; `waitForScan` itself can still block the first tool use. I kept the scope small here, but happy to follow up with a separate PR redesigning `waitForScan` to poll `is_scanning` internally if you prefer.

dmtrKovalenko: @gustav-fff please update the PR to include the refactoring and rebase on top of the main (use merge not rebase to be able to push to fork)

gustav-fff: [triage-bot] DIRECTED: cannot push merge to fork — gustav-fff OAuth token lacks `workflow` scope, and merging main brings in `.github/workflows/*` changes. Github rejects: `refusing to allow an OAuth App to create or update workflow .github/workflows/external-tests.yml without workflow scope`. Options: (1) you push the merge yourself from `cokekitten/fix-pi-fff-nonblocking-session-start`, or (2) re-target this PR to a same-repo branch so I can push. Also — need scope confirmation on the refactor: `packages/fff-node/src/finder.ts:438` already polls `isScanning()` async in JS; the blocking one is C FFI `fff_wait_for_scan` (`crates/fff-c/src/lib.rs:836`) used by fff-bun. Do you want the refactor in fff-node (drop the blocking C wait entirely, keep JS polling), in fff-bun (same async polling as fff-node), or in Rust `SharedFilePicker::wait_for_scan` (`crates/fff-core/src/shared.rs:119`)?

Honk-Honk 🪿

PR #485, opened 2026-05-18
Add "find_help_pages" function to search for help pages

SyntaxError2505:
I just copied the "find_files" function but made the directory be the file path of the documentation.

### DIFF
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -20,7 +20,8 @@ end
 function M.find_help_pages(opts)
   local picker_ok, picker_ui = pcall(require, 'fff.picker_ui')
   if picker_ok then
-    picker_ui.open({cwd = vim.env.VIMRUNTIME .. "/doc"})
+    local vimruntime = vim.env.VIMRUNTIME or vim.fn.stdpath('data')
+    picker_ui.open({cwd = vimruntime .. "/doc"})
   else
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
this is a good start, the one single problem is that it will reset the current index 

I am happy to merge this in but I need to get back to this allowing ceratin "one-shot" created picker instance that is destroyed on close. I think this is valuable overall! Would you like to work on something like this 

Also I think we should enforce modes to be { "Fuzzy" } by default becuase the directory is pretty small and fuzzy matching will just work better

SyntaxError2505:
Okay. I will fix the style errors and get to figuring that picker out. By fuzzy, do you mfn that live_grep like mode where you search for content or is there a function to search for filename?

dmtrKovalenko:
The mode for content search

SyntaxError2505:
I got to a point where I am happy. It is just a hidden find_files call with opts being passed trough but the path being set to the doc directory. As a user, the only thing I could make better is hide the icons because every file inside is just text anyways.

SyntaxError2505:
at this point it could also just be a keybind on user side

dmtrKovalenko:
well if this is just 3 lines function called the other public api does it need to even be a part of the api? Can we just documetn an example of being able to search help files like this from your custom config

SyntaxError2505:
That is what I mean. I'll definetly copy the code to my config. If you want to, you can put it into the README. It was fun to learn how plugins work tho

PR #487, opened 2026-05-19
fix(pi-fff): clarify tool prompt guidance

themauveavenger:
### Summary
The [official pi documentation](https://pi.dev/docs/latest/extensions#pi-registertool-definitionl) details how best to use `promptSnippet` and `promptGuidelines`. The three tools, `ffgrep`, `fffind`, and `fff-multi-grep` had these, but they weren't written in a way that an agent would be able to associate them with the actual tool. Specifically, each line in `promptGuidelines` should be written as "Use my_tool when..." because they are separated from where the tool's name and `promptSnippet` are loaded into the system prompt. 

The other change here is making it so that "@" doesn't end up as part of the query sent to `fff`. 

Validation:
- bun test packages/pi-fff/test/query.test.ts
- cd packages/pi-fff && bun run typecheck
- I've been using this extension as it is in pi for the last day with success and no errors. Agents have an easy time using the tools and don't fall back to bash as often.


dmtrKovalenko:
thanks but I need to find some time to run our eval layer

PR #496, opened 2026-05-20
fix: missing filename header when scrolling up in live_grep (#470)

### LINKED ISSUE #470: Top file name is missing
When I scroll all the way up in live_grep search that spans multiple pages, the filename for the top hit is missing. It is visible in the right pane (`validate.h`), which isn't present on the top.

Image

Neovim version: v0.12.2 (stable)
Fff.Nvim version: 1104a8deaf551fdb17fecadc87d6872397c98a6a

gustav-fff:
Closes #470

## Root cause

`lua/fff/grep/grep_renderer.lua:207` — file grouping logic uses `ctx._grep_last_file` to track consecutive matches from same file. This state resets per render (context is fresh each call). When user scrolls up past page boundary and new results load, first item in new page compared against `nil` → looks like different file but no way to distinguish "truly new file" from "first visible item after pagination". Result: no header rendered even though file might be same as last item on previous page.

## Fix

Always render header for first visible item (`ctx.iter_start`). Ensures filename always shown at top of viewport regardless of pagination state. Trade-off: duplicate header if scrolling within same file, but matches user expectation — "I see a match, I need to know which file".

## Steps to reproduce

Requires large repo with multi-file grep results spanning multiple pages:

\`\`\`sh
cd <large-repo>  # e.g. linux kernel, chromium, large monorepo
nvim
:lua require('fff').search('function')  -- common keyword with many hits
# Navigate down past first page boundary (e.g. j × 100)
# Navigate back up past page boundary (e.g. k × 100)
# Observe: top filename header missing on scrolled-to page
\`\`\`

Expected: filename header visible for topmost match.
Actual (pre-fix): filename missing, only `:line:col` and content shown.

## How verified

Code inspection: change targets exact root cause. Logic: `is_first_visible = (item_idx == ctx.iter_start)` ensures first rendered item always gets header. Checked signature match with `file_renderer.render_line` (third param already present but unused). Ran formatter (stylua passed).

Automated triage via gustav-fff bot.

### DIFF
--- a/lua/fff/picker_ui/grep_renderer.lua
+++ b/lua/fff/picker_ui/grep_renderer.lua
@@ -256,7 +256,11 @@ end
 ---@param ctx table Render context
 ---@return string[]
 function M.render_line(item, ctx)
-  local is_new_group = (item.relative_path ~= ctx._grep_last_file)
+  -- First rendered item in this pass always gets header — fixes missing header
+  -- when paginating backward in multi-page grep results (ctx is fresh per render).
+  local is_first_visible = not ctx._grep_first_rendered
+  ctx._grep_first_rendered = true
+  local is_new_group = is_first_visible or (item.relative_path ~= ctx._grep_last_file)
   ctx._grep_last_file = item.relative_path
 
   local match_line = render_match_line(item, ctx)
--- a/lua/fff/picker_ui/grep_renderer.lua
+++ b/lua/fff/picker_ui/grep_renderer.lua
@@ -254,12 +254,12 @@ end
 --- or 1 line [match] for subsequent matches in the same file.
 ---@param item FileItem Grep match item
 ---@param ctx table Render context
+---@param item_idx number 1-based item index in ctx.items
 ---@return string[]
-function M.render_line(item, ctx)
+function M.render_line(item, ctx, item_idx)
   -- First rendered item in this pass always gets header — fixes missing header
   -- when paginating backward in multi-page grep results (ctx is fresh per render).
-  local is_first_visible = not ctx._grep_first_rendered
-  ctx._grep_first_rendered = true
+  local is_first_visible = (item_idx == ctx.iter_start)
   local is_new_group = is_first_visible or (item.relative_path ~= ctx._grep_last_file)
   ctx._grep_last_file = item.relative_path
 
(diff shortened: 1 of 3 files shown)


dmtrKovalenko:
@gustav-fff rebase and retest thsi

gustav-fff:
[triage-bot] DIRECTED: rebased onto main (3a465e8).

Conflict in `lua/fff/picker_ui/grep_renderer.lua` — main moved the file from `lua/fff/grep/` and kept `render_line(item, ctx)` (no `item_idx` param). Resolved by tracking first-rendered via `ctx._grep_first_rendered` flag instead of comparing against `ctx.iter_start`. Same behavior, no signature change, list_renderer caller untouched.

CI will retest.

Honk-Honk 🪿

lua/fff/picker_ui/grep_renderer.lua
```diff
@@ -256,7 +256,11 @@ end
 ---@param ctx table Render context
 ---@return string[]
 function M.render_line(item, ctx)
-  local is_new_group = (item.relative_path ~= ctx._grep_last_file)
+  -- First rendered item in this pass always gets header — fixes missing header
+  -- when paginating backward in multi-page grep results (ctx is fresh per render).
+  local is_first_visible = not ctx._grep_first_rendered
```
dmtrKovalenko: why this is unused? 

keep this used and add this to the lua defintion params

cc @gustav-fff  work on this PR and babysit it

dmtrKovalenko (approved):

lua/fff/picker_ui/grep_renderer.lua
gustav-fff: [triage-bot] DIRECTED: pushed fbce94b.

`render_line(item, ctx, item_idx)` — `list_renderer` now passes `i` consistently with `apply_highlights`. First-visible check is `item_idx == ctx.iter_start`, no stateful flag. `file_renderer.render_line` signature updated to accept the optional param.

Honk-Honk 🪿

dmtrKovalenko:
@gustav-fff  _grep_last_file -> rename this to gre_last_file as it is now used 

AND ACUTALLY WAIT FOR CI TO PASS BEFORE LEAVING THIS PR

gustav-fff:
[triage-bot] DIRECTED: pushed 369a6a6 — renamed `ctx._grep_last_file` to `ctx.grep_last_file` (drop underscore prefix since the field is part of the render context now).

CI kicked off, currently pending across the build matrix. Will not consider this done until it goes green.

Honk-Honk 🪿

dmtrKovalenko:
@gustav-fff fix the luacheck lint

gustav-fff:
[triage-bot] DIRECTED: pushed ab5301c.

`lua/fff/picker_ui/file_renderer.lua:23` — added `-- luacheck: ignore item_idx` on the signature. The param is intentionally unused in `file_renderer` (only `grep_renderer` consumes it), keeping the shared `render_line(item, ctx, item_idx)` shape that `list_renderer` calls through.

Honk-Honk 🪿

PR #517, opened 2026-05-21
feat(picker): user mappings for input buffer (closes #78)

### LINKED ISSUE #78: [Feature Request] Allow mappings in insert mode
It would be nice if you could add mappings to the search input for insert mode, similarly to how telescope does it

```lua
                mappings = {
                    i = {
                        ["<A-BS>"] = function()
                            vim.api.nvim_input("<C-w>")
                        end
                    }
                },
```

In my case, I would like to remap `<A-BS>` (alt-backspace) to input `<C-w>`

gustav-fff:
Closes #78

## Root cause
No way for users to bind arbitrary lhs->callback on input buffer. Only named-action keymaps exposed via `keymaps` config.

## Fix
New `mappings` config table, telescope-style, keyed by mode (`i` / `n`). Values are function or string rhs. Applied after built-in keymaps in `lua/fff/picker_ui.lua:1198`, so user mappings override defaults for the same lhs.

```lua
require('fff').setup({
  mappings = {
    i = {
      ["<A-BS>"] = function() vim.api.nvim_input("<C-w>") end,
    },
  },
})
```

## Steps to reproduce
Pre-fix `main`: no `mappings` field; setting `<A-BS>` to send `<C-w>` is impossible without monkey-patching `picker_ui.setup_keymaps`.

After fix:

```lua
-- minimal repro init
require('fff').setup({
  mappings = {
    i = {
      ["<A-BS>"] = function() vim.api.nvim_input("<C-w>") end,
    },
  },
})
vim.keymap.set('n', '<leader>ff', require('fff').find_files)
```

Open picker, insert mode, type `foo bar`, press `<A-BS>` -> deletes `bar` (word back). Without fix, `<A-BS>` does nothing or inserts a literal char.

## How verified
- stylua format clean
- Manual: keymap registration loop iterates `i` and `n`, type-checks rhs, calls `vim.keymap.set` with input buffer opts. Built-ins set first so user mappings win on conflict.

Scope intentionally minimal per @dmtrKovalenko direction. Only input buffer for now (matches issue spec). Could extend to list/preview buffers later if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,14 @@ require('fff').setup({
     focus_list = '<leader>l',
     focus_preview = '<leader>p',
   },
+  -- Telescope-style user mappings on the input buffer, keyed by mode (`i` / `n`).
+  -- Values may be a function or a string (rhs). Applied on top of built-in keymaps.
+  mappings = {
+    i = {
+      -- ["<A-BS>"] = function() vim.api.nvim_input("<C-w>") end,
+    },
+    n = {},
+  },
   frecency = {
     enabled = true,
     db_path = vim.fn.stdpath('cache') .. '/fff_nvim',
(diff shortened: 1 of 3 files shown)


dmtrKovalenko:
🤦

PR #519, opened 2026-05-21
feat: add oldfiles picker for recently opened files

RicardoRamirezR:
## Summary

- Adds `find_oldfiles()` API and `:FFFOldFiles` command to browse recently opened files (`vim.v.oldfiles`)
- New `oldfiles` mode in the picker with Lua-side fuzzy filtering
- Reuses existing picker UI, file renderer, preview, and keymaps — zero Rust changes

## Motivation

Users coming from Telescope expect an oldfiles/recent files picker that works across projects. FFF's frecency system only tracks files within the current indexed directory, so files opened in other projects don't appear. This fills that gap.

## Implementation

- `lua/fff/oldfiles.lua` — reads `vim.v.oldfiles`, filters readable files, builds `FileItem`-compatible tables, fuzzy matches against query
- `lua/fff/main.lua` — new `M.find_oldfiles(opts)` entry point
- `lua/fff/picker_ui.lua` — new `'oldfiles'` branch in `update_results_sync()` (5 lines)
- `plugin/fff.lua` — `:FFFOldFiles` user command

## Usage

```lua
-- As keybinding
{ "fo", function() require("fff").find_oldfiles() end, desc = "Recent files" }

-- As command
:FFFOldFiles
```

## Test plan

- [x] Open files in different projects, verify they appear in oldfiles picker
- [x] Fuzzy search filters results correctly
- [x] File selection opens the correct file
- [x] Preview works for oldfiles items
- [x] Works with both `prompt_position = "top"` and `"bottom"`

### DIFF
--- a/lua/fff/main.lua
+++ b/lua/fff/main.lua
@@ -17,6 +17,23 @@ function M.find_files(opts)
   end
 end
 
+--- Find recently opened files (oldfiles)
+--- @param opts? table Optional configuration overrides
+function M.find_oldfiles(opts)
+  local picker_ok, picker_ui = pcall(require, 'fff.picker_ui')
+  if not picker_ok then
+    vim.notify('Failed to load picker UI: ' .. picker_ui, vim.log.levels.ERROR)
+    return
+  end
+
+  local picker_opts = vim.tbl_deep_extend('force', {
+    title = 'Recent Files',
+    mode = 'oldfiles',
+  }, opts or {})
+
+  picker_ui.open(picker_opts)
+end
+
 --- Live grep: search file contents in the current directory
 --- @param opts? {cwd?: string, title?: string, prompt?: string, layout?: table, grep?: {max_file_size?: number, smart_case?: boolean, max_matches_per_file?: number, modes?: string[]}, query?: string} Optional configuration overrides
 function M.live_grep(opts)
--- a/lua/fff/oldfiles.lua
+++ b/lua/fff/oldfiles.lua
@@ -1,0 +1,75 @@
+--- Oldfiles module for FFF
+--- Provides recent/old files list using vim.v.oldfiles as source
+local M = {}
+
+--- Build a FileItem-compatible table from a file path
+--- @param filepath string Absolute file path
+--- @param base_path string|nil Base path for computing relative_path
+--- @return table FileItem-compatible table
+local function build_file_item(filepath, base_path)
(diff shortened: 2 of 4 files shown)


dmtrKovalenko:
I am afraid I don't like this solution, I do not find much value in seeing the files opened in other projects? Why? if your cwd is something you opened right now what is the point of openning the files in the other directories

+ this makes the picker not actually use fff

RicardoRamirezR:
Fair point that this bypasses the Rust engine. The use case I'm solving: quickly opening files outside the cwd — temp files, downloads, global notes — without switching to another picker plugin. It keeps the workflow inside fff.

I understand if this doesn't fit the project's scope. Two options:

1. **Keep as-is** — lightweight Lua-only oldfiles, uses fff's picker UI but not the Rust engine
2. **Close** — if you'd rather not have features that don't go through the core

Happy to hear your preference.

dmtrKovalenko:
what is your use case tho? oldfiles is not a good interface for this - e.g. I have 1000 opened files in history and I am interested to see the file from the folder at 878. Or even worse - I know that I have a certain folder, but I have never opened it in vim yet. 

What is your ideal setup? e.g. I can think of having an ability to spawn one shot pickers that can open at any directory e.g. your nvim config, notes, other project, whatever (you just pass the dir for example) and dispose on close

RicardoRamirezR:
The advantage of oldfiles is discoverability — I don't need to know *where* the file is, just that I opened it recently. With directory-based pickers I'd have to preconfigure every location, and ad-hoc files (a temp file, something I downloaded once) would be missed.

That said, both features could coexist. Would you be open to supporting `find_files({ cwd = "~/notes" })` as a separate thing, and keeping oldfiles for the "I opened it recently but don't remember where" case?

dmtrKovalenko:
Conceptionally I see why it might make sense though realistically I don't want to have any features coming from the neovim side intermixed with the fff, so I'd rather say no to this feature addition at all. 

Though there is an interesting approach that both of us might be interested it in - a picker that uses not a vim's old files but it uses fff's access database to get the list of the most recently touched files across the whole file system and do it from there 

if you wish to prototype this API on the rust side we will definetely merge it, the current implemetnation that is not even using our internal matcher - won't be merged

RicardoRamirezR:
Thanks for the feedback and for considering this! I don't have the Rust expertise to implement the frecency-based approach you described, so I'll close this for now. Great project!

PR #532, opened 2026-05-25
feat(grep): jump to next/prev file group

### LINKED ISSUE #512: [FEATURE REQUEST] fff.nvim - Choose specific file in grep
Hi!
I am a frequent user of the grep functionality and sometimes there's a lot of results before the file I want to get to. 

It would be really nice if you could specify a specific file like `main.py`, and not only `*.py`.

I am a bit unsure how this could be implemented not to conflict with preexisting search modes, especially in the fuzzy, but perhaps in the plain search mode or a prefix to choose a specific file. 

A simple example shown below in a small repo. There are some ways to get around mashing arrow down or <C-n> but a built in way to grep a specific file or quickly choose the instance from the list would be really nice. 

Image

Image

gustav-fff:
Closes #512

## Root cause
Grep results are visually grouped by file, but no shortcut moves the cursor across group boundaries. Users with many matches in one file repeatedly mash `<Down>` / `<C-n>` to reach the next file's first hit.

## Fix
Add two grep-mode-only actions, `grep_jump_to_next_file` and `grep_jump_to_prev_file`, bound by default to `{<C-A-n>, <A-Down>}` and `{<C-A-p>, <A-Up>}`. They scan `filtered_items` from the current cursor for a `relative_path` change and jump to the first item of the new group. When no boundary remains on the current page, they call `load_next_page` / `load_previous_page` and place the cursor at the first match of the adjacent file in the new page. No-ops outside grep mode.

Both keymaps are configurable via `keymaps.grep_jump_to_next_file` / `keymaps.grep_jump_to_prev_file` (string or string[]). README + FffKeymapsConfig updated.

Diff: 3 source files + 1 test, ~102 LOC source.

## Steps to reproduce
Pre-fix `main`:
```sh
git checkout main
nvim --clean -u NONE
:lua vim.opt.runtimepath:prepend('/path/to/fff')
:lua require('fff').live_grep()
" type a query that matches many lines in one file (e.g. `function`)
" observe: only <Down>/<C-n> available; jumping past the first file requires N keypresses
```

Post-fix on this branch: `<C-A-n>` / `<A-Down>` jumps to first match of next file group; `<C-A-p>` / `<A-Up>` jumps backwards. Crosses page boundaries automatically.

## How verified
- `tests/grep_jump_file_spec.lua` (4 cases): forward jump across multiple groups, backward jump, no-op when not in grep mode, page-load when boundary lies on next page. All pass under `PlenaryBustedFile`.
- Manual: `nvim --headless -u tests/minimal_init.lua -c "PlenaryBustedFile tests/grep_jump_file_spec.lua"` -> `Success: 4, Failed: 0`.

Not exercised: `prompt_position='bottom'` rendering — purely cursor logic, render path unchanged. Maintainer should smoke-test in actual UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,9 @@ require('fff').setup({
     preview_scroll_down = '<C-d>',
     toggle_debug = '<F2>',
     cycle_grep_modes = '<S-Tab>',
+    -- grep mode only: jump cursor to first match of next/prev file group
+    grep_jump_to_next_file = { '<C-A-n>', '<A-Down>' },
+    grep_jump_to_prev_file = { '<C-A-p>', '<A-Up>' },
     cycle_previous_query = '<C-Up>',
     toggle_select = '<Tab>',
     send_to_quickfix = '<C-q>',
--- a/lua/fff/conf.lua
+++ b/lua/fff/conf.lua
@@ -35,6 +35,8 @@ local M = {}
 --- @field cycle_grep_modes string
 --- @field cycle_previous_query string
 --- @field cycle_forward_query string
+--- @field grep_jump_to_next_file string|string[]
+--- @field grep_jump_to_prev_file string|string[]
 --- @field toggle_select string
 --- @field send_to_quickfix string
 --- @field focus_list string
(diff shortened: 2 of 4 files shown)


dmtrKovalenko:
@gustav-fff rebase this PR to the new architecture

PR #544, opened 2026-05-31
feat(nvim): open all multi-selected files on cr

gshahbazian:
if multiple files are selected (using `<tab>`) then return should
open all of them. snacks picker works this way and i think most
will expect it.

### DIFF
--- a/tests/picker_dir_resolution_spec.lua
+++ b/tests/picker_dir_resolution_spec.lua
@@ -52,7 +52,7 @@ local function wait_for_scan(expected_dir, timeout_ms)
 end
 
 describe('picker find_files_in_dir path resolution (issue #389)', function()
-  local sandbox_root, target_dir, other_cwd, target_filename
+  local sandbox_root, target_dir, other_cwd, target_filename, second_filename
 
   before_each(function()
     sandbox_root = vim.fn.tempname()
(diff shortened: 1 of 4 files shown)


tests/picker_dir_resolution_spec.lua
```diff
@@ -61,9 +61,14 @@ describe('picker find_files_in_dir path resolution (issue #389)', function()
     vim.fn.mkdir(target_dir, 'p')
     vim.fn.mkdir(other_cwd, 'p')
 
-    target_filename = 'issue389_target.lua'
+    target_filename = 'issue389_target.txt'
```
dmtrKovalenko: please do not change any prior edited files for history completion

dmtrKovalenko (commented):
I have tried on snacks nvim and I don't see what they are doing it opens one of the match , and I am not sure what are you trying to do here.

Mayve you can upload the video of what you are performing? I think it might make more sense to open a new tab for every file (but I personally hate tabs)

gshahbazian:
@dmtrKovalenko this change opens a buffer for each multi-selected file. i will add a video for you when i get some time.

gshahbazian:
steps:
- open picker
- tab to select multiple files
- hit return

---

snacks picker

https://github.com/user-attachments/assets/b63a619a-4437-4ca6-bad7-4641cef3e72d

opens all selected

---

fff main

https://github.com/user-attachments/assets/bfcf394c-335f-4c5c-9319-052d584e7c2c

open only the first selection

---

fff this branch

https://github.com/user-attachments/assets/c6521c9a-ea0b-4a84-a5a9-03ecbf5de371

opens all selected

@dmtrKovalenko

PR #561, opened 2026-06-02
feat(cli): Add an rg-equivalent CLI interface for fff

markovejnovic:
This PR adds a CLI interface to `fff`. After talking with @dmtrKovalenko, we decided that the best architecture is to have an `fff` daemon run in the background to keep the index hot. The command that the user interacts with is `fff-rg`, which talks to this daemon.

Before we proceed, it's worth looking at the architecture diagram here:

```mermaid
flowchart LR
  subgraph fff-rg
    rg-searcher
    fffd-searcher
  end

  subgraph fff-daemon
    query-service
    subgraph session-pool
      session-1
      session-2
    end

    query-service --> session-1
    query-service --> session-2
  end

  rg-bin[[rg]]

  fffd-searcher <-->|unix sock|query-service
  rg-searcher <-->|subproc| rg-bin
```

Let's walk over each component:

### `fff-rg`

`fff-rg` can search either through talking through the `fff-daemon`, or through shelling to `rg`. It picks the `fff-daemon` if it believes to be working within a git repo. Otherwise, it will fall back to `rg`, and if the user doesn't have it installed, it will abort.

The core reasoning for this is that holding an `fff-daemon` alive for directories that don't need to be indexed makes no sense, as you'll hold a very large amount of RAM in memory for an effectively one-off search.

### `fff-daemon`

The `fff-daemon` holds within it a `query-service`. This `query-service` is responsible for accepting connections from new clients and then passing the request down to the `session-pool`.

The `session-pool` is a pool of "active" sessions, ie. active git repositories for which we have an `fff` index in memory. A couple notable facts:
- The session pool automatically reaps old sessions. If a search hasn't been performed for a while on a directory, chances are the user no longer cares for it, so we should evict it.
- The session pool has an LRU policy which means that if we're out of available pool slots, we'll kick out some unused and old session in favor of creating a new one.

### IPC protocol

I spent a good bit of time thinking through how best to handle the IPC protocol, and the epiphany I had is that we don't actually need to shuttle results between the daemon and the client. There may be hundreds, if not thousands of strings we'd need to copy, so if we can avoid it, that would be awesome.

On UNIX, you can pass file descriptors by using [SCM_RIGHTS](https://blog.cloudflare.com/know-your-scm_rights/) between processes. This enables us to take the stdout fd of the client, pass it to the daemon, and have the daemon directly write to the client's stdout. Neat!

The downside of this approach is that `SCM_RIGHTS` requires a unix domain socket, which means that passing the fd needs to happen over that. I evaluated:
- Adding a sidecar `iceoryx2` for all the other message passing, but ultimately I couldn't justify the maintenance burden. There's a ton of complexity added in figuring synchronizing the req-rep flow of the `mmap`ed iceoryx2 channels and the req-rep of the fd transfer, so it made no sense.
- Just having the req-rep flow between the client and the daemon live on the UDS (unix domain socket), which is ultimately what I landed on.

This lives in the `crates/cli/fff-ipc-domain` crate.

## Holes in the implementation

- Tests are quite lackluster. I added _some_ testing, but it's quite weak.
- There are a couple open questions I've left as comments on this PR.
- The LRU logic could be separated out to a separate container which is responsible for automatic eviction.

### DIFF
--- a/crates/cli/fff-daemon/Cargo.toml
+++ b/crates/cli/fff-daemon/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "fff-daemon"
+version = "0.0.1"
+edition = "2024"
+description = "FFF file finder daemon — indexes files and serves search requests over UDS."
+license = "MIT"
+
+[[bin]]
+name = "fff-daemon"
+path = "src/main.rs"
+
+[dependencies]
+fff = { package = "fff-search", path = "../../fff-core", version = "0.8.4" }
+fff-ipc-domain = { path = "../fff-ipc-domain" }
+rkyv = { workspace = true }
+clap = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+mimalloc = { workspace = true }
+parking_lot = { workspace = true }
+rayon = { workspace = true }
+smallvec = { workspace = true }
+dunce = { workspace = true }
+sendfd = "0.4"
+signal-hook = "0.3"
+
+[lints]
+workspace = true
--- a/crates/cli/fff-daemon/Cargo.toml
+++ b/crates/cli/fff-daemon/Cargo.toml
@@ -23,6 +23,3 @@ smallvec = { workspace = true }
 dunce = { workspace = true }
 sendfd = "0.4"
 signal-hook = "0.3"
-
-[lints]
-workspace = true
--- a/crates/cli/fff-daemon/src/convert.rs
+++ b/crates/cli/fff-daemon/src/convert.rs
@@ -1,0 +1,18 @@
+/// Bridges IPC domain types to their fff-core equivalents (orphan rule workaround).
+pub trait IntoCoreExt {
+    type Core;
+    /// Converts this IPC type into the corresponding fff-core type.
+    fn into_core(self) -> Self::Core;
+}
+
+impl IntoCoreExt for fff_ipc_domain::CaseMode {
+    type Core = fff::CaseMode;
+
+    fn into_core(self) -> fff::CaseMode {
+        match self {
+            Self::Smart => fff::CaseMode::Smart,
+            Self::Sensitive => fff::CaseMode::Sensitive,
+            Self::Insensitive => fff::CaseMode::Insensitive,
+        }
+    }
+}
--- a/crates/cli/fff-daemon/src/main.rs
+++ b/crates/cli/fff-daemon/src/main.rs
@@ -1,0 +1,59 @@
+//! Long-running daemon that indexes directories and serves search requests.
+//!
+//! The daemon binds a per-user Unix socket at `/tmp/fff-daemon-<uid>.sock`.
+//! Clients (typically `fff-rg`) connect and send a single request per
+//! connection:
+//!
+//! 1. Client sends a [`RequestHeader`] + rkyv-serialized [`SearchRequest`]
+//!    body, along with an output fd passed via SCM_RIGHTS.
+//! 2. Daemon looks up (or creates) a [`FilePicker`] for the requested
+//!    directory, runs the search, and writes results directly to the
+//!    client's output fd.
+//! 3. Daemon writes back a one-byte [`SearchStatus`] and closes the
+//!    connection.
+//!
+//! Directories are indexed on first request and kept alive in a session pool
+//! with a background file watcher. Subsequent queries against the same
+//! directory reuse the warm index.
+
+mod convert;
+pub(crate) mod output;
+mod query_service;
+mod session_pool;
+
+use std::sync::Arc;
+use std::sync::atomic::AtomicBool;
+
+use clap::Parser;
+
+use crate::query_service::QueryService;
+
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
+#[derive(Parser)]
+#[command(name = "fff-daemon", about = "FFF file finder daemon")]
+struct Args {
+    #[arg(long, default_value = "info", env = "FFF_LOG")]
+    log_level: String,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+                format!("fff_daemon={lvl},fff={lvl}", lvl = args.log_level).into()
+            }),
+        )
+        .init();
+
+    let shutdown = Arc::new(AtomicBool::new(false));
+    signal_hook::flag::register(signal_hook::consts::SIGTERM, shutdown.clone())
+        .expect("failed to register SIGTERM handler");
+    signal_hook::flag::register(signal_hook::consts::SIGINT, shutdown.clone())
+        .expect("failed to register SIGINT handler");
+
+    QueryService::new(shutdown).run();
+}
--- a/crates/cli/fff-daemon/src/query_service.rs
+++ b/crates/cli/fff-daemon/src/query_service.rs
@@ -1,0 +1,263 @@
+//! Listens on a per-user Unix domain socket for search requests. Each client
+//! connects, sends a [`RequestHeader`] + rkyv body alongside an output fd
+//! (via SCM_RIGHTS), and receives a single [`SearchStatus`] byte back.
+
+use std::io::{Read, Write};
+use std::num::NonZeroU64;
+use std::os::fd::OwnedFd;
+use std::os::unix::fs::PermissionsExt;
+use std::os::unix::io::FromRawFd;
+use std::os::unix::net::UnixListener;
+use std::path::Path;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
+
+/// Max time to wait for the client to send the request body.
+const READ_TIMEOUT: Duration = Duration::from_secs(30);
+/// Max time to wait for the file index to finish its initial scan.
+const SCAN_TIMEOUT: Duration = Duration::from_secs(30);
+/// Accept loop poll interval (non-blocking listener).
+const ACCEPT_POLL: Duration = Duration::from_millis(1);
+/// Owner-only permissions for the daemon socket.
+const SOCKET_MODE: u32 = 0o600;
+/// Inline buffer for small rkyv bodies — avoids heap allocation for typical requests.
+const IPC_BODY_INLINE: usize = 512;
+/// Default max file size for grep when the client doesn't specify one (4 MiB).
+const DEFAULT_MAX_FILE_SIZE: u64 = 4 * 1024 * 1024;
+
+use fff::{
+    FilePicker, FuzzySearchOptions, GrepMode, GrepSearchOptions, QueryParser, parse_grep_query,
+};
+use fff_ipc_domain::{
+    GrepQuery, GrepSearch, OutputFlags, RequestHeader, SearchKind, SearchRequest, SearchStatus,
+    daemon_socket_path,
+};
+use sendfd::RecvWithFd;
+
+use crate::convert::IntoCoreExt;
+
+use crate::session_pool::{MAX_SESSIONS, SessionPool};
+
+/// Parsed client request with the fd to write results into.
+struct IncomingQuery {
+    request: SearchRequest,
+    output: std::fs::File,
+}
+
+impl IncomingQuery {
+    /// Reads the 4-byte length header, receives the output fd, and deserializes the rkyv body.
+    fn recv(
+        stream: &mut std::os::unix::net::UnixStream,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        let mut raw_header = [0u8; RequestHeader::SIZE];
+        let mut fds = [0; 1];
+        let (n, fd_count) = stream.recv_with_fd(&mut raw_header, &mut fds)?;
+
+        if fd_count == 0 {
+            return Err("no fd in ancillary data".into());
+        }
+
+        // SAFETY: fd received via SCM_RIGHTS — this process owns it exclusively.
+        let owned_fd = unsafe { OwnedFd::from_raw_fd(fds[0]) };
+
+        if n < RequestHeader::SIZE {
+            return Err("incomplete header".into());
+        }
+
+        let header = RequestHeader::decode(raw_header);
+        let body_len = header.body_len as usize;
+
+        stream.set_read_timeout(Some(READ_TIMEOUT))?;
+        let mut body = smallvec::SmallVec::<[u8; IPC_BODY_INLINE]>::from_elem(0, body_len);
+        stream.read_exact(&mut body)?;
+
+        let archived = rkyv::access::<rkyv::Archived<SearchRequest>, rkyv::rancor::Error>(&body)
+            .map_err(|e| format!("rkyv access: {e}"))?;
+        let request = rkyv::deserialize::<SearchRequest, rkyv::rancor::Error>(archived)
+            .map_err(|e| format!("rkyv deserialize: {e}"))?;
+
+        Ok(Self { request, output: std::fs::File::from(owned_fd) })
+    }
+}
+
+/// RAII guard for the daemon's Unix listener socket. Cleans up the socket file on drop.
+struct ActiveDaemonSocket {
+    path: std::path::PathBuf,
+    listener: UnixListener,
+}
+
+impl ActiveDaemonSocket {
+    /// Binds a non-blocking Unix listener at the daemon socket path with 0600 permissions.
+    fn bind() -> Result<Self, Box<dyn std::error::Error>> {
+        let path = daemon_socket_path();
+        let _ = std::fs::remove_file(&path);
+
+        if let Some(parent) = path.parent() {
+            let _ = std::fs::create_dir_all(parent);
+        }
+
+        let listener = UnixListener::bind(&path)?;
+        let _ = std::fs::set_permissions(&path, std::fs::Permissions::from_mode(SOCKET_MODE));
+        listener.set_nonblocking(true).expect("failed to set listener non-blocking");
+
+        Ok(Self { path, listener })
+    }
+}
+
+impl Drop for ActiveDaemonSocket {
+    fn drop(&mut self) {
+        let _ = std::fs::remove_file(&self.path);
+    }
+}
+
+/// Accepts connections on the daemon socket and dispatches searches to worker threads.
+pub struct QueryService {
+    pool: Arc<SessionPool>,
+    shutdown: Arc<AtomicBool>,
+    workers: rayon::ThreadPool,
+}
+
+impl QueryService {
+    /// Creates the session pool and a rayon thread pool sized to `MAX_SESSIONS`.
+    pub fn new(shutdown: Arc<AtomicBool>) -> Self {
+        let pool = SessionPool::new();
+        let workers = rayon::ThreadPoolBuilder::new()
+            .num_threads(MAX_SESSIONS)
+            .thread_name(|i| format!("conn-{i}"))
+            .build()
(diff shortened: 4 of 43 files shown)


crates/cli/fff-ipc-domain/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Whether the client wants a fuzzy file search or a content grep.
+#[derive(Archive, Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[rkyv(derive(Debug))]
+pub enum SearchMode {
+    Files,
+    Grep,
+}
+
+/// Case sensitivity strategy for grep searches. Mirrors `fff::CaseMode` but
+/// with rkyv derives — fff-core doesn't depend on rkyv.
```
markovejnovic, line 23: this is up for debate whether we want to keep it this way or not.

I wanted to avoid adding another dependency to `fff` that is only used in the ipc code-path, but with optimizers being what they are, maybe that's not so bad.

either way, i chose this path of having two `CaseMode`s (one in `fff` and one in `fff-ipc-domain`), but open to discussion

crates/cli/fff-ipc-domain/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Output in `file:line:col:text` format for editor integration.
+    pub vimgrep: bool,
+}
+
+/// Full search request sent by the client as the rkyv body.
+#[derive(Archive, Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[rkyv(derive(Debug))]
+pub struct SearchRequest {
+    /// Root directory to search in (must be an absolute path).
+    pub directory: String,
```
markovejnovic: need to stop passing paths as strings. much better bet is to pass pathbufs as deep as possible until the fff barrier.

crates/fff-nvim/benches/grep_bench.rs
```diff
@@ -72,45 +71,34 @@ fn setup() -> &'static TestData {
...
-                watch: false,
-                ..Default::default()
-            },
-        )
-        .expect("create picker");
-
-        shared_picker.wait_for_scan(Duration::from_secs(120));
-        let file_count = {
-            let guard = shared_picker.read().expect("read lock");
-            guard.as_ref().expect("picker present").get_files().len()
```
dmtrKovalenko, line 101: this would require a major version bump

crates/cli/fff-rg/src/searcher/fffd.rs
```diff
@@ -0,0 +1,194 @@
+//! Client-side daemon searcher. Connects to the `fff-daemon` Unix socket,
+//! sends a search request with stdout as the output fd, and reads back a
+//! status byte. Spawns the daemon on first use if it isn't already running.
+
+use std::io::{Read, Write};
+use std::os::unix::io::AsRawFd;
```
dmtrKovalenko, line 6: ideally we should get all the unix specific mode into a separate file cause I would love this to work on windows at some point

crates/cli/fff-daemon/src/convert.rs
```diff
@@ -0,0 +1,18 @@
+/// Bridges IPC domain types to their fff-core equivalents (orphan rule workaround).
+pub trait IntoCoreExt {
+    type Core;
+    /// Converts this IPC type into the corresponding fff-core type.
+    fn into_core(self) -> Self::Core;
+}
+
+impl IntoCoreExt for fff_ipc_domain::CaseMode {
```
dmtrKovalenko, line 8: use From trait

crates/cli/fff-daemon/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+#[command(name = "fff-daemon", about = "FFF file finder daemon")]
+struct Args {
+    #[arg(long, default_value = "info", env = "FFF_LOG")]
+    log_level: String,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    tracing_subscriber::fmt()
```
dmtrKovalenko, line 44: use our log crate

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+/// Accept loop poll interval (non-blocking listener).
+const ACCEPT_POLL: Duration = Duration::from_millis(1);
+/// Owner-only permissions for the daemon socket.
+const SOCKET_MODE: u32 = 0o600;
+/// Inline buffer for small rkyv bodies — avoids heap allocation for typical requests.
+const IPC_BODY_INLINE: usize = 512;
+/// Default max file size for grep when the client doesn't specify one (4 MiB).
+const DEFAULT_MAX_FILE_SIZE: u64 = 4 * 1024 * 1024;
+
+use fff::{
```
dmtrKovalenko, line 29: do not intermix imports and constants

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+/// Max time to wait for the file index to finish its initial scan.
+const SCAN_TIMEOUT: Duration = Duration::from_secs(30);
+/// Accept loop poll interval (non-blocking listener).
+const ACCEPT_POLL: Duration = Duration::from_millis(1);
+/// Owner-only permissions for the daemon socket.
+const SOCKET_MODE: u32 = 0o600;
+/// Inline buffer for small rkyv bodies — avoids heap allocation for typical requests.
+const IPC_BODY_INLINE: usize = 512;
+/// Default max file size for grep when the client doesn't specify one (4 MiB).
+const DEFAULT_MAX_FILE_SIZE: u64 = 4 * 1024 * 1024;
```
dmtrKovalenko, line 27: should be 10

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+        let path = daemon_socket_path();
+        let _ = std::fs::remove_file(&path);
+
+        if let Some(parent) = path.parent() {
+            let _ = std::fs::create_dir_all(parent);
+        }
+
+        let listener = UnixListener::bind(&path)?;
+        let _ = std::fs::set_permissions(&path, std::fs::Permissions::from_mode(SOCKET_MODE));
+        listener.set_nonblocking(true).expect("failed to set listener non-blocking");
```
dmtrKovalenko, line 102: no panics, propagate the erro

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+struct IncomingQuery {
+    request: SearchRequest,
+    output: std::fs::File,
+}
+
+impl IncomingQuery {
+    /// Reads the 4-byte length header, receives the output fd, and deserializes the rkyv body.
+    fn recv(
+        stream: &mut std::os::unix::net::UnixStream,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
```
dmtrKovalenko, line 52: do not use dyn Error, use thiserror like our error.rs

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        if n < RequestHeader::SIZE {
+            return Err("incomplete header".into());
+        }
+
+        let header = RequestHeader::decode(raw_header);
+        let body_len = header.body_len as usize;
+
+        stream.set_read_timeout(Some(READ_TIMEOUT))?;
+        let mut body = smallvec::SmallVec::<[u8; IPC_BODY_INLINE]>::from_elem(0, body_len);
```
dmtrKovalenko, line 72: I don't think smallvec is efficient for 512 * 8 bytes it is going to be anyway allocated on the heap

use vec::with_capacity()

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+        stream.read_exact(&mut body)?;
```
dmtrKovalenko, line 73: I am unsure how we are understanding that this is enough bytes for the message?

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+    workers: rayon::ThreadPool,
+}
+
+impl QueryService {
+    /// Creates the session pool and a rayon thread pool sized to `MAX_SESSIONS`.
+    pub fn new(shutdown: Arc<AtomicBool>) -> Self {
+        let pool = SessionPool::new();
+        let workers = rayon::ThreadPoolBuilder::new()
+            .num_threads(MAX_SESSIONS)
+            .thread_name(|i| format!("conn-{i}"))
```
dmtrKovalenko, line 127: use fff- for thread names

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+            Err(e) => {
+                tracing::error!(err = %e, "failed to bind unix socket");
+                return;
+            }
+        };
+
+        tracing::info!(path = %socket.path.display(), "query service listening");
+
+        while !self.shutdown.load(Ordering::Relaxed) {
+            match socket.listener.accept() {
```
dmtrKovalenko, line 147: doese this mean that we can have a single conection?

crates/cli/fff-daemon/src/query_service.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /// Runs a fuzzy file search and writes formatted results to the output fd.
+    fn write_file_results(
+        picker: &FilePicker,
+        query: &str,
+        output: OutputFlags,
+        out: &mut std::fs::File,
+    ) -> Result<bool, Box<dyn std::error::Error>> {
+        let parsed = QueryParser::default().parse(query);
```
dmtrKovalenko, line 257: let's keep query parser on self to reuse it

crates/cli/fff-daemon/src/session_pool.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::sync::{Arc, Condvar, Mutex};
+use std::thread::JoinHandle;
+use std::time::{Duration, Instant};
+
+use parking_lot::RwLock;
+
+use fff::{FFFMode, FilePicker, FilePickerOptions, SharedFilePicker, SharedFrecency};
+
+/// Max concurrent directory indexes the daemon will keep alive.
+pub const MAX_SESSIONS: usize = 8;
```
dmtrKovalenko, line 12: all of these constants should become a default configuration

crates/cli/fff-daemon/src/session_pool.rs
```diff
@@ -0,0 +1,160 @@
...
+/// Sessions idle longer than this are evicted.
+const IDLE_TIMEOUT: Duration = Duration::from_mins(10);
+/// How often the evictor thread checks for idle sessions.
+const EVICTION_INTERVAL: Duration = Duration::from_secs(60);
+
+struct Session {
+    picker: SharedFilePicker,
+    #[allow(dead_code)]
```
dmtrKovalenko: ?

crates/cli/fff-daemon/src/session_pool.rs
```diff
@@ -0,0 +1,160 @@
...
+        if *guard {
+            return true;
+        }
+        let (guard, _) = self.cv.wait_timeout(guard, dur).unwrap();
+        *guard
+    }
+}
+
+pub struct SessionPool {
+    inner: Arc<Inner>,
```
dmtrKovalenko, line 55: i don't like this inner struct - whhy do we have it under Arc? is it changing? why can't we inline it as Arc<RwLock<..>>

crates/cli/fff-daemon/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition = "2024"
+description = "FFF file finder daemon — indexes files and serves search requests over UDS."
+license = "MIT"
+
+[[bin]]
+name = "fff-daemon"
+path = "src/main.rs"
+
+[dependencies]
+fff = { package = "fff-search", path = "../../fff-core", version = "0.8.4" }
```
dmtrKovalenko, line 13: use workspace dependencies

crates/cli/fff-daemon/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+tracing-subscriber = { workspace = true }
+mimalloc = { workspace = true }
+parking_lot = { workspace = true }
+rayon = { workspace = true }
+smallvec = { workspace = true }
+dunce = { workspace = true }
+sendfd = "0.4"
+signal-hook = "0.3"
+
+[lints]
```
dmtrKovalenko: remove

crates/cli/fff-ipc-domain/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+//! All structured payloads use rkyv zero-copy serialization. The socket path
+//! is per-user via [`daemon_socket_path`].
+
+use std::num::{NonZeroU32, NonZeroU64};
+use std::path::PathBuf;
+
+use rkyv::{Archive, Deserialize, Serialize};
+
+/// Returns the daemon socket path: `/tmp/fff-daemon-<uid>.sock`.
+pub fn daemon_socket_path() -> PathBuf {
```
dmtrKovalenko, line 17: this should also be a part of configuration

crates/cli/fff-ipc-domain/src/lib.rs
dmtrKovalenko: we can have optional dependencies - this is absolutely fine

crates/cli/fff-ipc-domain/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Always case-sensitive.
+    Sensitive,
+    /// Always case-insensitive.
+    Insensitive,
+}
+
+/// Grep pattern — either a regex or a literal string.
+#[derive(Archive, Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[rkyv(derive(Debug))]
+pub enum GrepQuery {
```
dmtrKovalenko, line 38: hmmm I would've structured as mode: Regex, query: tring, and leave a space for additional parametrs - e.g. page size and offset 

we should also figure out if we want to have proper pagination here at somet point

crates/cli/fff-ipc-domain/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Literal string match.
+    Literal(String),
+}
+
+/// Grep-specific search parameters.
+#[derive(Archive, Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[rkyv(derive(Debug))]
+pub struct GrepSearch {
+    /// The search pattern.
+    pub query: GrepQuery,
```
dmtrKovalenko, line 50: oh I see - ignore previous comment, just add mode from fff direclty

we can add optional dependency to fff and we anyway need rkyv soon for another feature

crates/cli/fff-ipc-domain/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+pub enum SearchStatus {
+    /// At least one result was found and written to the output fd.
+    Match = 0,
+    /// Search completed successfully but produced no results.
+    NoMatch = 1,
+    /// Search failed (e.g. indexing timeout, invalid query).
+    Failed = 2,
+}
+
+impl From<SearchStatus> for u8 {
```
dmtrKovalenko, line 135: use as_u8 functin

crates/cli/fff-rg/src/searcher/fffd.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        Command::new(bin)
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .process_group(0)
+            .spawn()
+            .map_err(|e| format!("failed to spawn {}: {e}", bin.display()))?;
+
+        // Instant::now() is vDSO/commpage — no syscall; connect() dominates.
```
dmtrKovalenko, line 99: remove this comment or make it more clear

crates/cli/fff-rg/src/types/cli/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+
+#[derive(Parser)]
+#[command(
+    name = "fff-rg",
+    about = "FFF — daemon-accelerated file finder and grep",
+    after_help = "Falls back to rg when searching outside a git repository."
+)]
+/// Mirrors a subset of `rg` flags so `fff-rg` is a drop-in replacement.
+#[allow(clippy::struct_excessive_bools)]
+pub struct Args {
```
dmtrKovalenko, line 43: i think we should restructure the crates to be somethjing like this

cli/
  daemon
  ffd
  frg
  ipc 


this iwll make it much easier to keep backward compatibility with those tools

crates/cli/fff-rg/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+mod types;
+
+use clap::Parser;
+
+use crate::app_ctx::AppCtx;
+use crate::searcher::{Search, Searcher};
+use crate::types::cli::Args;
+
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
```
dmtrKovalenko, line 23: i don't think we need mimaloc here - can drop a dependency

crates/cli/fff-rg/tests/rg_compat/synthetic.rs
```diff
@@ -0,0 +1,147 @@
+use crate::util::Dir;
+
+const EXTENSIONS: &[&str] = &["rs", "ts", "json", "md", "txt", "toml", "yaml"];
```
dmtrKovalenko, line 3: for tests I would prefer tests in the big repo using proptest on various flags + queries

crates/cli/fff-rg/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+fff-ipc-domain = { path = "../fff-ipc-domain" }
+rkyv = { workspace = true }
+sendfd = "0.4"
+clap = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+bytesize = "2"
+mimalloc = { workspace = true }
+git2 = { workspace = true }
+which = "8.0.3"
```
dmtrKovalenko, line 22: im not sure we need it - if it's on a path Command will resolve it

crates/cli/rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+edition = "2024"
```
dmtrKovalenko, line 1: pls do not override

dmtrKovalenko (commented):
some nits on the code I need to go through the way it actually works one more time

crates/cli/fff-daemon/src/session_pool.rs
markovejnovic: oops, slop that passed through

dmtrKovalenko:
@gustav-fff please split out all the changes related to the new case_sensetivity api to a separate PR (keep Co-Authored-by marco) but make sure that we are not doing any breaking changes

gustav-fff:
[triage-bot] DIRECTED: split case-sensitivity API into #609.

Approach: added `CaseMode` enum + `case_mode: Option<CaseMode>` field on `GrepSearchOptions` alongside the existing `smart_case: bool`. Old field kept; when `case_mode` is `None` it falls back to the boolean (`true` => Smart, `false` => Sensitive). Identical behavior for existing callers — no breaking change to public structs or function signatures. `cargo check -p fff-search` clean.

Marko credited via Co-authored-by trailer on the commit.

Once #609 merges, this PR can drop the `case_mode: ...` lines that re-translate `smart_case` and just set `case_mode: Some(...)` directly when Insensitive is needed.

Honk-Honk 🪿

PR #565, opened 2026-06-03
feat(nvim): Make `move_up` and `move_down` work in normal mode.

mkblast:
Sorry, It was so annoying I couldn't help myself...


mkblast:
Also, I don't want to make an issue for this question, but is it feasible to make FFF work with multiple paths?Like make it accept an array of paths and searches/greps in them? Or is it something beyond the project?

dmtrKovalenko:
it is possible, there is no limitation on how many pikcers you can have though each of those will have it's own resource pull but it definetely can be done right

dmtrKovalenko (approved):

Issue #403, opened 2026-04-21
pi-fff: parallel ffgrep calls can hang indefinitely due to unsynchronized ensureFinder singleton


0xbentang:
### Summary

When the pi `ffgrep` tool is called in parallel (≥2 concurrent tool calls in a single turn), the shared `ensureFinder` singleton can deadlock or corrupt state, causing the calls to hang with no `toolResult` returned.

### Steps to reproduce

1. Start a pi session in a large monorepo.
2. Fire ≥2 parallel `ffgrep` tool calls in the same assistant turn.
3. All calls enter `ensureFinder()` simultaneously. At least one caller hangs indefinitely.

### Expected behavior

All parallel `ffgrep` calls return within a few seconds.

### Actual behavior

Tool calls hang with no result. The TUI freezes. The session must be killed manually.

### Root cause

`packages/pi-fff/src/index.ts` line 278–298 — `ensureFinder` has no mutex or in-flight promise cache. Concurrent callers overwrite the shared `finder` variable while another caller is still awaiting `waitForScan` on a stale reference. The native `FileFinder.create` path also acquires exclusive SQLite locks on the frecency/history DBs — two concurrent creates can deadlock at the native layer with no timeout.

### Suggested fix

Serialize with an in-flight promise cache:

```ts
let finderPromise: Promise<FileFinder> | null = null;

function ensureFinder(cwd: string): Promise<FileFinder> {
  if (finder && !finder.isDestroyed && finderCwd === cwd) return Promise.resolve(finder);
  if (finderPromise) return finderPromise;
  finderPromise = (async () => {
    if (finder && !finder.isDestroyed) finder.destroy();
    const result = FileFinder.create({ basePath: cwd, frecencyDbPath, historyDbPath, aiMode: true });
    if (!result.ok) throw new Error(result.error);
    finder = result.value;
    finderCwd = cwd;
    await finder.waitForScan(15000);
    return finder;
  })().finally(() => { finderPromise = null; });
  return finderPromise;
}
```

### Environment

- `@ff-labs/pi-fff` version: **0.6.0** (same code on `main` as of 2026-04-21)
- `pi` version: latest (installed via npm)
- Working tree: large monorepo (~500k files)

### Workaround until fixed

Instruct the agent to serialize `ffgrep`/`fffind`/`fff-multi-grep` calls rather than issuing them in parallel.

PR #570, opened 2026-06-04
Use composable autocomplete provider instead of replacing the editor

LinYS77:
## Summary

   This PR changes `pi-fff` to provide its FFF-backed `@` file mentions through Pi's composable autocomplete provider API instead of replacing the
 editor component.

   The current implementation installs a custom editor with `ctx.ui.setEditorComponent(...)` only to customize autocomplete behavior. That gives
 `pi-fff` ownership of the whole editor surface, even though the extension only needs to contribute file mention suggestions.

   Using `ctx.ui.addAutocompleteProvider(...)` keeps the same `@` mention functionality while allowing `pi-fff` to compose cleanly with other editor
 extensions.

   ## Changes

   - Remove the custom `FffEditor` wrapper.
   - Stop calling `ctx.ui.setEditorComponent(...)`.
   - Register FFF `@` mentions with `ctx.ui.addAutocompleteProvider(...)`.
   - Delegate non-FFF autocomplete behavior to the existing provider.
   - Keep mode/tool behavior unchanged.

   ## Testing

   - `cd packages/pi-fff && bun test test/`
   - `cd packages/fff-node && npm run build`
   - `git diff --check`

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -1,26 +1,25 @@
 /**
  * pi-fff: FFF-powered file search extension for pi
  *
- * Overrides built-in `find` and `grep` tools with FFF and can also replace
- * @-mention autocomplete suggestions in the interactive editor.
+ * Overrides built-in `find` and `grep` tools with FFF and adds FFF-backed
+ * @-mention autocomplete suggestions to the interactive editor.
  */
 
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
-import { CustomEditor } from "@earendil-works/pi-coding-agent";
 import {
-  Text,
   type AutocompleteItem,
   type AutocompleteProvider,
+  Text,
 } from "@earendil-works/pi-tui";
-import { Type } from "@sinclair/typebox";
-import { FileFinder } from "@ff-labs/fff-node";
 import type {
   GrepCursor,
   GrepMode,
   GrepResult,
-  SearchResult,
   MixedItem,
+  SearchResult,
 } from "@ff-labs/fff-node";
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
@SamuelLHuber do you have any objections?

dmtrKovalenko:
@gustav-fff review this PR

SamuelLHuber:
looks good reading it. will test the branch now

SamuelLHuber (approved):
good one. tested and verified locally on my machine too

gustav-fff:
[triage-bot] DIRECTED: LGTM.

Composition via `addAutocompleteProvider((current) => ...)` is correct. Mention path tries first, falls through to `current` on null/abort/throw. `shouldEnableMentions()` evaluated per-call so `/fff-mode` toggles take effect without re-registration — the removed `applyEditorMode` re-apply on mode change is intentional and safe.

Notes (non-blocking):
- `applyCompletion` always delegates to `current`. Works because `item.value` already encodes `@"path with space"`/`@path`, so prefix-length replace produces the right result. If a future base provider gets clever about prefix handling (e.g. unquoting), mention insertion could regress — worth a single inline test that asserts the inserted line for an `@`-mention.
- `catch {}` in `getSuggestions` swallows native errors silently. Acceptable for autocomplete UX, but a `pi.notify`/debug log on first failure would help diagnose `FileFinder` startup races (#403-class issues).
- `shouldTriggerFileCompletion` defaults to `true` when base lacks it — fine for mentions, but could double-fire base file completion alongside `@` results in some hosts. Watch in real session.

Tests pass: `bun test test/` → 17 pass.

Honk-Honk 🪿

Issue #389, opened 2026-04-18
Bug: find_files_in_dir doesn't preview or open with the correct path


estevesnp:
When I use `find_files_in_dir` (or even run `live_grep` with a different `cwd`), the files are properly populated in the picker from the provided directory.

However, the preview doesn't show the actual file contents (shows `No preview available` instead), and when the file is opened, it doesn't open the actual file (it tries to open it relatively to the dir where neovim was opened).

For example, running `require("fff").find_files_in_dir(vim.fn.stdpath("config"))` while in my home dir:

<img width="589" height="1396" alt="Image" src="https://github.com/user-attachments/assets/0be79cb2-5db8-4943-8a94-1c5fffa7eff6" />

estevesnp:
was fixed by 71c33e63b203dde0c7995e7db2a1c613d7e90f34, will close issue

PR #578, opened 2026-06-06
feat(ui): expose select.pre_select_hook to override window retargeting

### LINKED ISSUE #577: fff.nvim doesnt always open buffers in the window i invoked it from
There's a bug for me where if I have a split tab in neovim (two side-by-side buffers), and one buffer is a file view (eg `:Ex` or `:Oil`) and the other buffer is an opened file, if I open fff from the file explorer buffer and then select a file, it opens that file in the buffer with a file already opened. I expect it to open the buffer from the window I opened it in, just as it does when both of the open buffers are a file. 

I have only observed this behavior when one buffer is a file and the other buffer is a directory view (both `:Ex` and `:Oil` reproduce this) and I open fff from the window with the directory viewer.

gustav-fff:
Closes #577

## Root cause

`picker_ui/picker_ui.lua` hard-codes the rule: if the invoking window has a special `buftype`, is non-modifiable, or has `winfixbuf`, retarget the open to a different window via `find_suitable_window()`. `oil` and `:Ex` buffers trip the `buftype` check, so picking a file from them lands in the *other* split instead of replacing the dir-view, contradicting the reporter's mental model from telescope (`:e <file>` in the invoking window).

## Fix

Lift the retargeting decision into a user-overridable hook, per the design agreed in #288 (incl. @dmtrKovalenko's instruction to keep all existing checks intact in the default).

- New config: `select.pre_select_hook(current_buf, action)` where `action` is `'edit' | 'split' | 'vsplit' | 'tab'`.
- Default hook reproduces the prior behavior verbatim (special buftype / non-modifiable / `winfixbuf` -> `find_suitable_window()`).
- `M.select()` calls the hook before issuing `:edit`/`:split`/`:vsplit`/`:tabedit`.
- `winfixbuf` `:split` fallback is kept as a hard guard against `E1513` regardless of hook outcome.
- Dead-code: removed the local `find_suitable_window` wrapper in `picker_ui.lua` since `M.close()` already tears down picker floats before the hook runs, so float-exclusion isn't needed.

To get telescope-style \"always open in invoking window\" (the reporter's ask), users override with a no-op:

```lua
require('fff').setup({
  select = {
    pre_select_hook = function(_buf, _action) end,
  },
})
```

Documented in `README.md` under \"Open in invoking window\".

## Steps to reproduce

```sh
git clone https://github.com/dmtrKovalenko/fff /tmp/fff-577 && cd /tmp/fff-577
git checkout main
make build
mkdir -p /tmp/repro577 && cd /tmp/repro577
echo 'one' > a.txt && echo 'two' > b.txt
nvim --clean -u NONE -c "set rtp+=/tmp/fff-577" -c "lua require('fff').setup({})" \
  -c "edit a.txt" -c "vsplit" -c "Ex" -c "wincmd h"
```

Trigger from the netrw window:
- `:lua require('fff').find_files()`
- type `b.txt`, press `<CR>`

Expected (per reporter): `b.txt` opens in the netrw window (replacing it).
Actual on `main`: `b.txt` opens in the right-hand window where `a.txt` was, netrw stays.

## How verified

- Default behavior: unchanged. The new default hook executes the same buftype/modifiable/winfixbuf branch the inline code did.
- Opt-in fix: with `pre_select_hook = function() end` set, the netrw window is the one `:edit b.txt` runs in, matching telescope semantics.
- `winfixbuf` regression guard: `M.select` still detects `winfixbuf` post-hook and falls back to `:split`, so `E1513` cannot escape.
- `luac -p` clean. `make format` (cargo + stylua) clean. TS biome step skipped — not installed locally and no TS files touched.
- No rust changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,18 @@ require('fff').setup({
   git = {
     status_text_color = false, -- true to color filenames by git status
   },
+  select = {
+    -- Runs right before the picker opens the chosen file (after the picker
+    -- has closed). Default retargets the open to a "suitable" window when
+    -- the invoking window has a special buftype, is non-modifiable, or has
+    -- 'winfixbuf' set — so picking from oil/Ex/snacks-explorer/outline
+    -- doesn't replace the sidebar with the file.
+    --   current_buf: the buffer that was active when <CR> was pressed
+    --   action:      'edit' | 'split' | 'vsplit' | 'tab'
+    -- Override to force-open in the invoking window — see "Open in invoking
+    -- window" below.
+    pre_select_hook = function(current_buf, action) --[[ default impl ]] end,
+  },
   grep = {
     max_file_size = 10 * 1024 * 1024,
     max_matches_per_file = 100,
@@ -381,6 +393,20 @@ Grep-only:
 - `src/main.rs`. Grep inside a single file.
 
 Mix freely: `git:modified src/**/*.rs !src/**/mod.rs user controller`.
+
+### Open in invoking window
+
+By default the picker retargets the open to a different window when invoked from a non-file buffer (oil, `:Ex`, sidebars). To force-open in whichever window invoked the picker — telescope-style `:edit <file>` semantics — set `select.pre_select_hook` to a no-op:
+
+```lua
+require('fff').setup({
+  select = {
+    pre_select_hook = function(_current_buf, _action) end,
+  },
+})
+```
+
+Caveat: the chosen file replaces the buffer in the invoking window even if it's a non-modifiable / special buftype. `winfixbuf` windows still fall back to `:split` to avoid `E1513`.
 
 ### Multi-select and quickfix
 
--- a/README.md
+++ b/README.md
@@ -321,16 +321,10 @@ require('fff').setup({
     status_text_color = false, -- true to color filenames by git status
   },
   select = {
-    -- Runs right before the picker opens the chosen file (after the picker
-    -- has closed). Default retargets the open to a "suitable" window when
-    -- the invoking window has a special buftype, is non-modifiable, or has
-    -- 'winfixbuf' set — so picking from oil/Ex/snacks-explorer/outline
-    -- doesn't replace the sidebar with the file.
-    --   current_buf: the buffer that was active when <CR> was pressed
-    --   action:      'edit' | 'split' | 'vsplit' | 'tab'
-    -- Override to force-open in the invoking window — see "Open in invoking
-    -- window" below.
-    pre_select_hook = function(current_buf, action) --[[ default impl ]] end,
+    -- Return winid to open the chosen file in, or nil to open in the
+    -- invoking window. Default retargets when the invoking window can't
+    -- host a file buffer (oil/:Ex/sidebars, non-modifiable, winfixbuf).
+    select_window = function(current_buf, action) --[[ default impl ]] end,
   },
   grep = {
     max_file_size = 10 * 1024 * 1024,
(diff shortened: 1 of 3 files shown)


lua/fff/conf.lua
```diff
@@ -349,6 +355,25 @@ local function init()
...
     },
+    select = {
+      --- Runs right before the picker opens the chosen file. Reposition the
+      --- cursor to a different window, abort the open via error(), or no-op.
+      --- Default retargets when the current window can't host a file buffer
+      --- (special buftype, non-modifiable, or winfixbuf) — set to a no-op
+      --- function to always open in the invoking window.
+      --- @param current_buf integer
+      --- @param action FffSelectAction
+      pre_select_hook = function(current_buf, action)
```
dmtrKovalenko: let's name this function select_window and make it return the winid, if nil returned - it should simply open in place in the previous window

dmtrKovalenko (approved):
@gustav-fff addresss comments and keep comments more terse

lua/fff/conf.lua
gustav-fff: [triage-bot] DIRECTED: renamed `pre_select_hook` -> `select_window`. Returns `integer|nil`: nil = open in invoking window, winid = switch first. Default still retargets on special buftype/non-modifiable/winfixbuf. README + types updated. Pushed `73d2136`.

Honk-Honk 🪿

dmtrKovalenko:
@gustav-fff babysit the CI until its green

README.md
```diff
@@ -382,6 +388,20 @@ Grep-only:
 
 Mix freely: `git:modified src/**/*.rs !src/**/mod.rs user controller`.
 
+### Open in invoking window
+
+By default the picker retargets the open to a different window when invoked from a non-file buffer (oil, `:Ex`, sidebars). To force-open in whichever window invoked the picker — telescope-style `:edit <file>` semantics — return `nil` from `select.select_window`:
```
dmtrKovalenko:
```suggestion
By default fff.nvim will try to open a file in the most suitable window, so any non-file buffers are not affected. You can customize or disable this by providing:
```

README.md
```diff
@@ -320,6 +320,12 @@ require('fff').setup({
   git = {
     status_text_color = false, -- true to color filenames by git status
   },
+  select = {
+    -- Return winid to open the chosen file in, or nil to open in the
+    -- invoking window. Default retargets when the invoking window can't
+    -- host a file buffer (oil/:Ex/sidebars, non-modifiable, winfixbuf).
```
dmtrKovalenko:
```suggestion
    -- Return winid to open the chosen file in, or nil to open in the original window
```

lua/fff/conf.lua
```diff
@@ -349,6 +355,23 @@ local function init()
...
+    select = {
+      --- Returns winid to open the file in. Return nil to open in the invoking
+      --- window. Default retargets when the invoking window can't host a file
+      --- buffer (special buftype, non-modifiable, or winfixbuf).
+      --- @param current_buf integer
+      --- @param action FffSelectAction
+      --- @return integer|nil
+      select_window = function(current_buf, action)
+        if action ~= 'edit' then return nil end
+        local current_win = vim.api.nvim_get_current_win()
```
dmtrKovalenko, line 367: can you doublecheck if this will actually keep the current work by doing the actual neovim instance testing?

dmtrKovalenko:
@gustav-fff work on this PR

gustav-fff:
[triage-bot] DIRECTED: pushed 9c25162.

Two CI failures had same root cause: `M.close()` runs before `vim.schedule` callback, resets `M.state.config` to nil. Deferred lookup of `config.select.select_window` crashed:

```
.../lua/fff/picker_ui/picker_ui.lua:324: attempt to index local 'config' (a nil value)
```

That's why `picker_dir_resolution_spec.lua` (#389 regression test) failed on `e2e (ubuntu/windows)`. Same call site triggered the luacheck shadowing warning on line 354.

Fix: capture `config = M.state.config` alongside `location/query/mode/suggestion_source` before `M.close()`. Renamed inner `config = conf.get()` to `cfg` to clear the shadow.

Applied both README suggestions verbatim (lines 326, 393).

Verified locally on macos:
- `make test-lua` — full suite green, including `picker_dir_resolution_spec.lua` which was the failing test on CI.
- `luacheck lua/` — `0 warnings / 0 errors in 36 files`.
- `stylua --check` — clean.

Re your verification ask on `lua/fff/conf.lua:367`: the failing CI test IS the real-nvim verification you asked for. `picker_dir_resolution_spec.lua` drives the same `picker_ui.select('edit')` code path that `<CR>` hits — populates state, calls select, waits for the scheduled `:edit`, asserts the buffer is the fixture from `target_dir`. Pre-fix it crashed inside the new hook lookup. Post-fix it passes. So the default `select_window` semantics (retarget on special buftype/non-modifiable/winfixbuf) are exercised end-to-end against a real Neovim.

Honk-Honk 🪿

dmtrKovalenko (approved):

PR #583, opened 2026-06-08
fix(nvim): keep picker movement responsive

JonasThowsen:
I ran into an issue in Neovim where quickly scrolling through picker results with `<C-n>` / `<C-p>` could make the list freeze until I released the key. In some cases it also produced the error shown below:

screenshot-from-08-06-2026 12-34-58

This PR improves cursor movement responsiveness by avoiding a full list redraw when moving between results. Instead, for renderers that support it, it only updates the old and new cursor rows.

It also debounces preview updates during navigation so holding movement keys does not immediately trigger preview work for every cursor step. I used `10ms` because it felt responsive in local testing while still reducing repeated preview updates, but I’m happy to adjust that value if there’s a preferred threshold.

Custom renderers fall back to the full render path unless they explicitly opt into the cursor-row optimization, so renderers that change row text based on cursor state keep working correctly.

### DIFF
--- a/lua/fff/file_picker/preview.lua
+++ b/lua/fff/file_picker/preview.lua
@@ -28,13 +28,15 @@ local function append_buffer_lines(bufnr, lines)
   vim.api.nvim_set_option_value('modifiable', false, { buf = bufnr })
 end
 
+local function show_preview_unavailable(bufnr) set_buffer_lines(bufnr, { 'No preview available' }) end
+
 local function set_preview_filetype(bufnr, filetype)
-  -- Preserve user-facing errors from preview-local filetype churn.
+  -- FileType hooks can throw or mutate global v:errmsg during preview setup.
+  -- pcall lets us restore it; callers treat failures as preview-local.
   local errmsg = vim.v.errmsg
-  vim.v.errmsg = ''
   local ok, err = pcall(vim.api.nvim_set_option_value, 'filetype', filetype, { buf = bufnr })
   vim.v.errmsg = errmsg
-  if not ok then error(err) end
+  return ok, err
 end
 
 local function find_existing_buffer(file_path)
@@ -289,7 +291,10 @@ local function link_buffer_content(source_bufnr, target_bufnr)
   set_buffer_lines(target_bufnr, lines)
 
   local source_ft = vim.api.nvim_get_option_value('filetype', { buf = source_bufnr })
-  if source_ft ~= '' then set_preview_filetype(target_bufnr, source_ft) end
+  if source_ft ~= '' then
+    local ok = set_preview_filetype(target_bufnr, source_ft)
+    if not ok then return false end
+  end
 
   M.state.has_more_content = false
   M.state.total_file_lines = #lines
@@ -445,7 +450,11 @@ function M.preview_file(file_path, bufnr)
       set_buffer_lines(bufnr, content)
 
       local file_config = M.get_file_config(file_path)
-      set_preview_filetype(bufnr, info.filetype)
+      local ok = set_preview_filetype(bufnr, info.filetype)
+      if not ok then
+        show_preview_unavailable(bufnr)
+        return
+      end
       vim.api.nvim_set_option_value('modifiable', false, { buf = bufnr })
       vim.api.nvim_set_option_value('readonly', true, { buf = bufnr })
       vim.api.nvim_set_option_value('buftype', 'nofile', { buf = bufnr })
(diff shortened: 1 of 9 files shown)


lua/fff/file_picker/preview.lua
```diff
@@ -87,25 +96,27 @@ local function init_dynamic_loading_async(file_path, callback)
...
 
   vim.uv.fs_open(file_path, 'r', 438, function(err, fd)
-    -- Stale callback: preview moved on to a different file
-    if M.state.preview_generation ~= generation then
-      if fd then pcall(vim.uv.fs_close, fd) end
-      return
-    end
+    vim.schedule(function()
+      -- Stale callback: preview moved on to a different file
+      if M.state.preview_generation ~= generation then
```
dmtrKovalenko, line 103: this is actaully so slow with snacks.nvim I have a plan to rewrite this for so long

lua/fff/picker_ui/renderer.lua
```diff
@@ -274,4 +278,50 @@ function M.render_list()
   finalize_render(separator_line, ctx)
 end
 
+local function rerender_cursor_rows(old_cursor, new_cursor)
+  if S.suggestion_source then return false end
+  if not (S.list_buf and vim.api.nvim_buf_is_valid(S.list_buf)) then return false end
+  if not (S.list_win and vim.api.nvim_win_is_valid(S.list_win)) then return false end
+  if not (S.item_to_lines and S.last_render_ctx and S.ns_id) then return false end
```
dmtrKovalenko: do we really need those checks?

lua/fff/file_picker/preview.lua
```diff
@@ -28,6 +28,15 @@ local function append_buffer_lines(bufnr, lines)
   vim.api.nvim_set_option_value('modifiable', false, { buf = bufnr })
 end
 
+local function set_preview_filetype(bufnr, filetype)
+  -- Preserve user-facing errors from preview-local filetype churn.
+  local errmsg = vim.v.errmsg
+  vim.v.errmsg = ''
+  local ok, err = pcall(vim.api.nvim_set_option_value, 'filetype', filetype, { buf = bufnr })
```
dmtrKovalenko, line 37: can you write a comment here why we are doing this a pall? what might go wrong here and why we are doing ignore and then propatating the error back? 

if this we want to preserve the vim.v.errmsg why it is not safe to simply ignore the error and simply stop rendering the preview?

lua/fff/picker_ui/grep_renderer.lua
```diff
@@ -7,6 +7,8 @@ local fuzzy = require('fff.fuzzy')
 local file_renderer = require('fff.picker_ui.file_renderer')
 local tresitter_highlight = require('fff.treesitter_hl')
 
+M.supports_cursor_rerender = true
```
dmtrKovalenko: put this inline to the M = { } defiention

tests/picker_ui_snap.lua
```diff
@@ -41,6 +41,7 @@ local function setup(geometry, opts)
         vim.opt.runtimepath:prepend(plugin)
         package.path = plugin .. '/lua/?.lua;' .. plugin .. '/lua/?/init.lua;' .. package.path
         vim.cmd('cd ' .. vim.fn.fnameescape(%q))
+        vim.cmd('syntax off')
```
dmtrKovalenko, line 44: what's the point of doing this in tests?

dmtrKovalenko (approved):
Generally I see the idea and I support it, just had a few questions/nits

tests/picker_ui_snap.lua
JonasThowsen: I had some undeterministic output for the snapshot test that was caused by the syntax highlighting changing some attr values. Adding the syntax off made the test predictable, and I thought it made sense as it seemed like this should only test the picker UI. If that is not the case, then this is probably the wrong approach.

lua/fff/picker_ui/renderer.lua
JonasThowsen: Probably overly defensive, will address that in a new commit

lua/fff/file_picker/preview.lua
JonasThowsen: Good point. The pcall here is not meant to hide a real picker error, it is mostly used as a small finally block. Setting filetype on the preview buffer can run user/plugin FileType hooks, and those hooks may throw or overwrite global v:errmsg. Since this happens only while setting up a scratch preview buffer, the failure should stay preview-local: restore the previous v:errmsg, abort this preview, and keep the picker usable. I changed it to return ok, err instead of rethrowing, so callers can stop rendering the preview instead of propagating the error.

PR #595, opened 2026-06-12
feat(grep): Jump to next file match segment

### LINKED ISSUE #512: [FEATURE REQUEST] fff.nvim - Choose specific file in grep

Hex4C:
Fully closes #512 

---

Heavily inspired by #532 (squashed based on that PR)

Since a majority of the logic already was implemented and only needed to be slightly adapted to the new structure for the picker_ui I wrote a majority of the code with LLM assisted tools.

I did the same thing for the unit test, but I am less certain of them... They seemed slightly dated, they can almost certainly be improved.

I have tested the UI for edge cases such as no files, 1 file and multiple files etc and it has worked flawlessly so far.

Feedback is greatly appreciated or edit the files directly if you wish. Hope this helps!

### DIFF
--- a/lua/fff/picker_ui/navigation.lua
+++ b/lua/fff/picker_ui/navigation.lua
@@ -196,4 +196,94 @@ function M.scroll_preview_down()
   preview.scroll(scroll_lines)
 end
 
+-- Helper function to eliminate UI update redundancy
+local function update_ui_after_jump(old_cursor)
+  if not P.render_after_cursor_move(old_cursor) then return false end
+  P.update_status()
+  pcall(vim.cmd, 'redraw')
+  P.update_preview_debounced()
+  return true
+end
+
+function M.grep_jump_to_next_file()
+  if not P.state.active or S.mode ~= 'grep' then return end
+  local items = S.filtered_items
+  if not items or #items == 0 then return end
+
+  local old_cursor = S.cursor
+  local current_path = items[S.cursor] and items[S.cursor].relative_path
+
+  for i = S.cursor + 1, #items do
+    if items[i].relative_path ~= current_path then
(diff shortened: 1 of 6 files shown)


Hex4C:
Force pushed to remove picker_ui.lua (original before refactor), accidentally added it.

Hex4C:
Messed up the test setup, fixed now and passes the 4 unit tests.

lua/fff/picker_ui/navigation.lua
```diff
@@ -117,7 +117,7 @@ function M.move_up()
 
   if not P.render_after_cursor_move(old_cursor) then return end
   P.update_status()
-  pcall(vim.cmd, 'redraw')
```
dmtrKovalenko, line 120: why did you remove pcalls? They are there on purpose

Hex4C: Whoops my bad, had some lsp warnings and wanted to use vim.cmd.redraw to avoid the type mismatch warning. Meant to use *pcall*(vim.cmd.redraw), not sure though if it is exactly the same so I'll leave as it was before (with pcall(vim.cmd, 'redraw'))

Issue #603, opened 2026-06-15
SIGSEGV in notify-rs debouncer loop during Node SDK file watching on macOS arm64


ChengShiLiu16:
## Summary

`@ff-labs/fff-node` can intermittently crash the hosting Node process with `SIGSEGV` / `EXC_BAD_ACCESS` on macOS arm64 while the background file watcher is enabled.

The captured macOS crash report shows the faulting thread is named `notify-rs debouncer loop`. The crash is inside `_platform_memmove`, called from the bundled `libfff_c.dylib`, with `KERN_INVALID_ADDRESS at 0x0000000000000010`.

Because this happens on a native FFI worker thread, the embedding JS app does not receive a normal JS exception or run its usual teardown path.

## Environment

- Frontend: Node SDK (`@ff-labs/fff-node`)
- `@ff-labs/fff-node`: `0.8.4` in the crashing app
- Native binary: `@ff-labs/fff-bin-darwin-arm64/libfff_c.dylib`
- Bundled watcher crates observed in the dylib: `notify-9.0.0-rc.4`, `fff-notify-debouncer-full-0.9.4`
- Latest checked npm packages: `@ff-labs/fff-node@0.9.4` and `@ff-labs/fff-bin-darwin-arm64@0.9.5-nightly.ec57eb0` still contain the same watcher crate versions
- OS: macOS 27.0 build 26A5353q
- CPU: Apple Silicon / arm64
- Node: v22.22.0
- `FileFinder.create({ basePath, frecencyDbPath, historyDbPath, aiMode: true })`
- Default watcher behavior: watcher enabled (`disableWatch` not set)

## Crash report excerpt

```text
Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: 0x0000000000000001, 0x0000000000000010
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000000000000010

Thread 25 (CRASHED) — name: "notify-rs debouncer loop"
  libsystem_platform.dylib   _platform_memmove
  libfff_c.dylib             +0x7e34c
  libfff_c.dylib             +0x630c4
  libfff_c.dylib             +0xcfb08
  libfff_c.dylib             +0x99498
  libfff_c.dylib             +0xa8abc
  libfff_c.dylib             +0x20ad30
  libsystem_pthread.dylib    _pthread_start
  libsystem_pthread.dylib    thread_start

Related watcher threads alive at crash:
[...]

PR #604, opened 2026-06-15
test(node): add watcher churn stress target

ChengShiLiu16:
## Summary

Adds a focused Node SDK watcher churn stress target for the macOS FSEvents / debouncer crash reported in #603.

This does not claim to fix the crash. It gives maintainers a closer repro/diagnostic loop for the failing path: Node SDK + watcher enabled + concurrent filesystem churn + concurrent searches.

## Changes

- Add `packages/fff-node/test/watcher-churn-stress.mjs`
  - creates a temp watched tree
  - starts `FileFinder` with the watcher enabled
  - waits for scan and watcher readiness
  - mutates files with create/write/rename/delete churn
  - concurrently runs file, directory, and grep searches
  - writes per-iteration fff logs under `/tmp`
- Add `make test-node-watcher-stress`
- Add `make build-c-lib-prof`, `make prepare-node-prof`, and `make test-node-watcher-stress-prof`
  - uses the existing `profile.prof` profile so macOS crashes can be reproduced with debuginfo/symbols retained

## Validation

Local:

- `node --check packages/fff-node/test/watcher-churn-stress.mjs`
- `npm ci --ignore-scripts`
- `cd packages/fff-node && npm run build`
- `git diff --check`
- `make -n test-node-watcher-stress-prof`

Not run locally:

- `make test-node-watcher-stress`
- `make test-node-watcher-stress-prof`

Reason: the local Rust toolchain currently has no usable `cargo` component and `rustup` rolls back while installing due to a `rust-analyzer` component conflict. The new Makefile targets expand correctly and should run in a working Rust environment / CI.

Refs #603

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ SHELL := bash
 # string rather than the literal `-o` / `pipefail` tokens.
 .SHELLFLAGS := -o pipefail -ec
 
-.PHONY: build build-c-lib install uninstall test test-rust test-c-smoke test-c-api test-lua test-lua-snap test-version test-bun test-node prepare-bun prepare-bun-packaged prepare-node set-npm-version header test-stress test-stress-seeded test-stress-random test-stress-repos test-node-stress sync-js-api sync-js-api-check bump-homebrew-formula bump-install-mcp-sh test-bun-compile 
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
image

not sure what you mean but this is not reproducible for me not for mac not on linux

Issue #247, opened 2026-02-18
API: more stracture and pass callback to all provided APIs


neo451:
https://github.com/obsidian-nvim/obsidian.nvim/issues/364 there's this issue in my repo that has been there a while, and I visit this repo from time to time to see if live grep has landed, and i found it has got it today, so I went ahead trying to integrate this plugin

but I do find the implementations a bit weird and very limited, I get the fact that this set out as a file picker, but passing a callback is just the most natural thing, and there's a `open_with_callback` but that is not what drives the main APIs like `find_files` and `live_grep`

API-wise, after not finding a clear API guide in the readme, I just grep these calls I want to make in your repo, but finding these are still not enough for my needs, since there's no really types in the `opts` param, just a `table`, I had to further trace the next level of functions that they calls. that really is not ideal for a repo that seems mature with many users.

and the type annotations I found is just simply wrong: https://github.com/dmtrKovalenko/fff.nvim/blob/9bc928db65be91a4c580d5ca8d1844ad9901d24c/lua/fff/picker_ui.lua#L2488-L2488

typing `opts` like this is just not valid type annotation for either lua-ls or emmylua-ls, it is just helping no one sitting there, seems too big of a mistake to have just sit around in such an active repo for 4 months

ok after this I found even more recent similar styled annotations, so maybe @dmtrKovalenko prefers this, but aren't you getting any diagnostics? and it is just not easy for others to call your code like this

dmtrKovalenko:
Oh yeah I don’t use any lua lsp and type diagnostics are just generated by LLM when I asked it to lool

I probably should revisit the way docs are done. But it’s unclear what is missing concurrently? I made open with callback to support one specific use case, what are you trying to do?

dmtrKovalenko:
https://github.com/dmtrKovalenko/fff.nvim/blob/9bc928db65be91a4c580d5ca8d1844ad9901d24c/lua/fff/main.lua

This is btw the main file

neo451:
this is how people integrate pickers, using obsidian.nvim's snacks integration as an example:

```lua
---@param opts obsidian.PickerFindOpts|? Options.
M.find_files = function(opts)
  opts = opts or {}
  opts.callback = opts.callback or obsidian.api.open_note

  ---@type obsidian.Path
  local dir = opts.dir.filename and Path.new(opts.dir.filename) or Obsidian.dir

[...]

PR #606, opened 2026-06-16
feat(picker_ui): add opts.on_submit hook for custom selection action

### LINKED ISSUE #605: [Suggestion]: A way to programmatically do something with the selection results from UI picker
### Which fff frontend(s)?

Neovim plugin (fff.nvim)

I want to integrate fff.nvim with oil.nvim so that after I get the result from `find_files()`, I can use the result to open the directory containing that file using oil.

As a reference, the following is the way to configure this using `snacks.picker` and `fzf.lua`:
```lua
--- snacks.picker
keys = {
  {
      '<leader>fo',
      function()
        Snacks.picker.files({
          title = 'Oil',
          -- override the confirm action
          confirm = function(picker, item)
            picker:close()
            if not item then
              return
            end
            local file_path = Snacks.picker.util.path(item) or item.file
            if not file_path then
              return
            end
            local dir_path = file_path:match('^(.*)[/\\][^/\\]*$')
            require('oil').toggle_float(dir_path)
          end,
        })
      end,
      desc = 'Find directory (Oil)',
  }
}
```
```lua
--- fzf.lua
keys = {
   {
      '<leader>fo',
      function()
        FzfLua.files({
          prompt = 'Oil❯ ',
          actions = {
            -- override the default action
            ['default'] = function(selected, opts)
              opts.cwd = opts.cwd or vim.uv.cwd()
              if selected and #selected > 0 then
                local path = vim.split(selected[1], ' ')[2]
                local file_path = opts.cwd .. '/' .. path
                local dir_path = file_path:match('^(.*)[/\\][^/\\]*$')
                require('oil').toggle_float(dir_path)
              end
            end,
          },
        })
      end,
      desc = 'Find directory (Oil)',
   }
}
```

### Proposed solution

I think the simplest non-breaking API change is by changing the public API of `file_search(query, opts)` so that the `query` argument becomes optional instead of required. When the `query` is `nil`, the UI is opened instead for the user to interact.

An alternative is by allowing custom actions in `opts` so that `find_files(opts)` can call a callback can be callback after the selection is done. But honestly, I don't know how to make the API nice. The `select.select_window` config seems to assumes that the chosen file must be opened in the buffer.

gustav-fff:
Closes #605

## Root cause
Picker UI hardcodes `:edit` (or `split`/`vsplit`/`tab`) on selection at `lua/fff/picker_ui/picker_ui.lua:498`. The only extension point — `select.select_window` (`lua/fff/conf.lua:65`) — picks the target window but cannot replace the open action itself. Use-cases like \"open the directory of the selected file in oil.nvim\" have no hook.

## Fix
Add `opts.on_submit` to `find_files(opts)`. When set, it runs in place of the default open action after the picker closes. Receives the item plus a context table (action, abs path, relative path, location, query, mode). Query history tracking still runs. No behavior change when `on_submit` is unset.

Signature: `fun(item, ctx)` where `ctx = { action, path, relative_path, location, query, mode }`.

## Steps to reproduce
Pre-fix on `main`: there is no way to intercept selection. After the patch:

```lua
require('fff').find_files({
  title = 'Oil',
  on_submit = function(item, ctx)
    local dir = ctx.path:match('^(.*)[/\\\\][^/\\\\]*$')
    require('oil').toggle_float(dir)
  end,
})
```

Expected: oil opens at the directory containing the selected file. Actual (pre-fix): the file is `:edit`-ed; reporter has no hook to override.

## How verified
Manual review of `M.select` flow: `on_submit` runs inside the same `vim.schedule` after `M.close()`, so picker teardown completes before the callback fires (same ordering as the default `:edit` path). When `on_submit` is unset the original path is taken verbatim.

Did not run a Neovim-level e2e — flagging for maintainer to sanity-check the oil integration interactively.

### DIFF
--- a/lua/fff/picker_ui/picker_ui.lua
+++ b/lua/fff/picker_ui/picker_ui.lua
@@ -483,9 +483,36 @@ function M.select(action)
   vim.cmd('stopinsert')
   M.close()
 
+  local on_submit = config and config.on_submit
+
   -- Defer file open past picker float teardown. Without this, foldexpr is not
   -- recomputed on the new window (folds appear missing) on some platforms.
   vim.schedule(function()
+    if type(on_submit) == 'function' then
+      local ok, err = pcall(on_submit, item, {
+        action = action,
+        path = abs_path,
+        relative_path = relative_path,
+        location = location,
+        query = query,
+        mode = mode,
+      })
+      if not ok then vim.notify('FFF: on_submit error: ' .. tostring(err), vim.log.levels.ERROR) end
+
+      if query and query ~= '' then
+        local cfg = config or conf.get()
+        if cfg.history and cfg.history.enabled then
+          local fff = require('fff.core').ensure_initialized()
+          if mode == 'grep' then
+            pcall(fff.track_grep_query, query)
+          else
+            pcall(fff.track_query_completion, query, item.relative_path)
+          end
+        end
+      end
+      return
+    end
+
     if config and config.select and type(config.select.select_window) == 'function' then
       local ok, win = pcall(config.select.select_window, vim.api.nvim_get_current_buf(), action)
       if not ok then
--- a/lua/fff/picker_ui/picker_ui.lua
+++ b/lua/fff/picker_ui/picker_ui.lua
@@ -498,49 +498,36 @@ function M.select(action)
         mode = mode,
       })
       if not ok then vim.notify('FFF: on_submit error: ' .. tostring(err), vim.log.levels.ERROR) end
-
-      if query and query ~= '' then
-        local cfg = config or conf.get()
-        if cfg.history and cfg.history.enabled then
-          local fff = require('fff.core').ensure_initialized()
-          if mode == 'grep' then
-            pcall(fff.track_grep_query, query)
-          else
-            pcall(fff.track_query_completion, query, item.relative_path)
-          end
+    else
+      if config and config.select and type(config.select.select_window) == 'function' then
+        local ok, win = pcall(config.select.select_window, vim.api.nvim_get_current_buf(), action)
+        if not ok then
+          vim.notify('FFF: select.select_window error: ' .. tostring(win), vim.log.levels.WARN)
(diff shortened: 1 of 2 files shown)


lua/fff/picker_ui/picker_ui.lua
```diff
@@ -483,9 +483,36 @@ function M.select(action)
...
+        action = action,
+        path = abs_path,
+        relative_path = relative_path,
+        location = location,
+        query = query,
+        mode = mode,
+      })
+      if not ok then vim.notify('FFF: on_submit error: ' .. tostring(err), vim.log.levels.ERROR) end
+
+      if query and query ~= '' then
```
dmtrKovalenko: @gustav-fff do not duplicate the behvior it is literally a duplicated code block. Also I am not sure should we expose some api to close the picker or we can rely that users are moving to the other window?

gustav-fff: [triage-bot] DIRECTED: pushed 1b05da9. Made `on_submit` an `else` branch over the default open path so query-history tracking is shared (single block at the end). No duplicated code now.

On exposing a close API: not needed for `on_submit` itself — `M.close()` already runs before the `vim.schedule` callback fires, so by the time the user hook executes the picker is gone. The hook just opens whatever window/buffer it wants. Happy to expose `require('fff').close()` as a public symbol if you want it for other reasons (e.g. external keymaps), but it's not required for this feature. Leaving it out for now.

Honk-Honk 🪿

dmtrKovalenko (approved):

dmtrKovalenko:
@gustav-fff I remember there was another related issue please find it and link

gustav-fff:
[triage-bot] DIRECTED: related issue is #247 ("API: more structure and pass callback to all provided APIs") — same theme of exposing callback hooks on the public API. This PR addresses the selection-callback slice; #247 is broader (callbacks across all APIs) and remains open.

Honk-Honk 🪿

PR #615, opened 2026-06-18
fix(nvim): keep prompt icon when clearing input with cc/S

### LINKED ISSUE #614: cc/S in the prompt (vim mode) leaves a stray prompt icon that must be backspaced
## Description

With `prompt_vim_mode = true`, clearing the query from normal mode with `cc` (or `S`) leaves a stray copy of the prompt icon (e.g. `🪿`) in the input bar that has to be manually backspaced out.

## Steps to reproduce

1. Set `prompt_vim_mode = true`.
2. Open the picker and type a query.
3. Press `<Esc>` to enter normal mode, then `cc` (or `S`).

**Expected:** the query is cleared and the prompt icon remains as the prefix.
**Actual:** the line is fully cleared and a literal prompt icon is left behind that must be backspaced.

## Root cause

The input bar is a `buftype=prompt` buffer with the prompt set via `prompt_setprompt` (`lua/fff/picker_ui/ui_creator.lua`). The prompt icon is protected prompt text at the start of the line. `cc`/`S` delete the entire line including the prompt, and the prompt buffer then re-inserts the icon as plain literal text.

KorigamiK:
## Summary

Fixes #614.

With `prompt_vim_mode = true`, clearing the query from normal mode with `cc`/`S` left a stray prompt icon behind that had to be backspaced out.

The input bar is a `buftype=prompt` buffer whose prompt icon is protected text at the start of the line. `cc`/`S` delete the whole line including the prompt, so Neovim re-inserts the icon as plain literal text.

## Fix

In `prompt_vim_mode`, map `cc` and `S` to reset the line to just the prompt and re-enter insert at the correct column. Clearing the query now leaves the prompt intact, and the resulting empty-query search runs as normal.

## Testing

- Manually verified `cc` and `S` from normal mode in `prompt_vim_mode` clear the query and keep the prompt icon, with the cursor placed right after it in insert mode.

### DIFF
--- a/lua/fff/picker_ui/ui_creator.lua
+++ b/lua/fff/picker_ui/ui_creator.lua
@@ -357,6 +357,22 @@ function M.setup_keymaps()
   if S.config.prompt_vim_mode then
     set_keymap('n', keymaps.close, P.close, input_opts)
     set_keymap('i', '<C-c>', P.close, input_opts)
+
+    -- cc/S clear the whole line, wiping the prompt buffer's prompt and leaving
+    -- a stray icon behind. Reset to just the prompt and re-enter insert instead.
+    local function clear_query_line()
+      local prompt = S.config.prompt
+      vim.api.nvim_set_option_value('modifiable', true, { buf = S.input_buf })
+      vim.api.nvim_buf_set_lines(S.input_buf, 0, -1, false, { prompt })
+      vim.schedule(function()
+        if S.input_win and vim.api.nvim_win_is_valid(S.input_win) then
+          vim.api.nvim_win_set_cursor(S.input_win, { 1, #prompt })
+          vim.cmd('startinsert!')
+        end
+      end)
+    end
+    set_keymap('n', 'cc', clear_query_line, input_opts)
+    set_keymap('n', 'S', clear_query_line, input_opts)
   else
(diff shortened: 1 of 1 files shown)


lua/fff/picker_ui/ui_creator.lua
```diff
@@ -357,6 +357,22 @@ function M.setup_keymaps()
...
+      vim.api.nvim_set_option_value('modifiable', true, { buf = S.input_buf })
+      vim.api.nvim_buf_set_lines(S.input_buf, 0, -1, false, { prompt })
+      vim.schedule(function()
+        if S.input_win and vim.api.nvim_win_is_valid(S.input_win) then
+          vim.api.nvim_win_set_cursor(S.input_win, { 1, #prompt })
+          vim.cmd('startinsert!')
+        end
+      end)
+    end
+    set_keymap('n', 'cc', clear_query_line, input_opts)
```
dmtrKovalenko: we probably need remap=true here to be able to remap existing bindings of cc

dmtrKovalenko (approved):

lua/fff/picker_ui/ui_creator.lua
KorigamiK: Cool, I've updated this for the cc/S mappings so existing user remaps still resolve.

PR #617, opened 2026-06-20
refactor(core): collapse SharedFrecency/SharedQueryTracker into generic SharedDb

seuros:
SharedFrecency and SharedQueryTracker are now type aliases over a single SharedDb<T: LmdbStore>, and the three wait_for_* methods share one poll_until helper. Public API unchanged; 

Sorry for the negative contribution :). -60 LOC

### DIFF
--- a/crates/fff-core/src/shared.rs
+++ b/crates/fff-core/src/shared.rs
@@ -2,7 +2,7 @@ use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard, Weak};
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
this is a hard breaking change - i don't think we want to make. They are exposed to the users of SDK and separate types to make users be able to choose if they want frecency, query tracking, or both.

dmtrKovalenko:
oh wait I actually read the code now lol 

this makes sense

seuros:
Ahaha, i was going to reply : `It's your OWN CODE`. And some comments...

I can open similar PR while i'm reading/understanding the code.

seuros:
@dmtrKovalenko Do you have an Eta for this PR ?

PR #533, opened 2026-05-27
feat: Optimize repeatable greps for large files

dmtrKovalenko:
better utilization of virtual memory for larger files to reduce amount of syscals we spent on chunked reading

experimentatlly set values when based on real hardware results

PR #626, opened 2026-06-22
fix: read grep content into an owned buffer to avoid SIGBUS on concurrent file modification

kwojcik-blockether:
## Problem

Grep can **SIGBUS and abort the entire host process** when a file is modified while it is being searched.

`FileItem::get_content_for_search` returns file content via **mmap** — a per-`FileItem` `OnceLock<Mmap>` cache, plus a fresh mmap for files `>= FRESH_MMAP_THRESHOLD` (introduced in #533). mmap-backed reads are unsound under concurrent modification: when an editor or a coding agent rewrites/truncates a file **between or during** greps — the normal case for the Neovim picker and the MCP agent — the mapping's pages past the new EOF become invalid, and the next read (including the SIMD prefilter `case_insensitive_memmem::search_packed_pair` scanning the whole file) faults with `SIGBUS`, killing the process. A `SIGBUS` from an invalid mapping cannot be caught from safe Rust, so there is no recovery.

The cached mapping makes it worse: an `OnceLock<Mmap>` created at index time can be served long after the file changed on disk.

This surfaced in production as:

```
SIGBUS (0xa) ...
C  [libfff_c.dylib]  fff_search::case_insensitive_memmem::search_packed_pair
C  [libfff_c.dylib]  fff_search::grep::grep_search
```

## Root cause, not the search kernel

To rule out a real over-read in the SIMD memmem, I added a **guard-page unit test** (`case_insensitive_memmem::tests::guard_page_no_overread`): it places haystacks ending exactly at a `PROT_NONE` page and fuzzes lengths 1..600 × needle 2..48. It passes — the kernel never reads past a valid slice. **The fault is the invalid mapping, not the search.**

## Fix

Read grep content into the existing per-worker scratch `Vec<u8>` instead of mmap-ing it. The buffer is reused across files (one bounded copy per file, no per-query allocation) and is immune to truncation — a shrinking file simply yields its new, shorter bytes.

## Reproduction (deterministic)

Added `grep_reflects_file_rewrite_without_stale_mmap`: index a ~2 MB file, grep it, rewrite it much shorter with the needle removed (no re-index), grep again.

- **Before:** the second grep returns the stale cached match, or — once the file shrinks past a page boundary — aborts with `process didn't exit successfully ... (signal: 10, SIGBUS: access to undefined memory)`.
- **After:** the grep correctly reflects the new content (0 matches).

## Tradeoff

Grep content is no longer zero-copy mmap. If the repeatable-large-file perf from #533 matters, the cache could instead hold an owned `Box<[u8]>` (read once, invalidated when the file's size/mtime changes) — same speedup, still crash-safe. Happy to follow up that way if you prefer.

## Tests

- `cargo test -p fff-search` — all green (94 lib + integration), including the new regression and guard-page tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/fff-core/src/types.rs
+++ b/crates/fff-core/src/types.rs
@@ -711,22 +711,21 @@ impl FileItem {
 
         // Read the file into an owned, reusable buffer rather than mmap it.
         //
-        // mmap-backed grep races with concurrent modification: if the file is
-        // truncated or rewritten while a mapping is live (an editor or coding
-        // agent changing the workspace), the pages past the new EOF become
-        // invalid and the next read — including a SIMD prefilter load over the
-        // whole file — faults with SIGBUS, aborting the entire process. The old
-        // cached fast path made it worse: an `OnceLock<Mmap>` created at index
-        // time could be served long after the file changed on disk. An owned
-        // read() copy is immune; a shrinking file just yields a shorter buffer,
-        // never a fault. `buf` is a per-worker scratch Vec reused across files,
-        // so this is one bounded copy per file, not a per-query allocation. (The
-        // content cache stays warm for the file picker via `get_cached_content`;
-        // only the grep reader stopped trusting a possibly-stale mapping.)
+        // Read into a reusable owned buffer, not mmap: an mmap'd file that is
+        // truncated/rewritten mid-grep (editor or agent edits) faults with
+        // SIGBUS on the pages past its new EOF. `buf` is per-worker, reused.
         let _ = mmap_slot;
-        let mut file = std::fs::File::open(&abs).ok()?;
+        let file = std::fs::File::open(&abs).ok()?;
         buf.clear();
-        file.read_to_end(buf).ok()?;
+        // Cap the read at the budget: `self.size` is the stale indexed size, so
+        // a file that GREW past `max_file_size` since indexing must not balloon
+        // memory. Read one byte over the cap and bail if it exceeds it.
+        file.take(max_file_size.saturating_add(1))
+            .read_to_end(buf)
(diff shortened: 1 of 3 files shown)


crates/fff-core/src/types.rs
```diff
@@ -702,38 +702,30 @@ impl FileItem {
...
-            let stored = mmap_slot.insert(mmap);
-            return Some(&stored[..]);
-        } else {
-            let _ = (mmap_slot, arena);
+        // Read the file into an owned, reusable buffer rather than mmap it.
+        //
+        // Read into a reusable owned buffer, not mmap: an mmap'd file that is
+        // truncated/rewritten mid-grep (editor or agent edits) faults with
+        // SIGBUS on the pages past its new EOF. `buf` is per-worker, reused.
+        let _ = mmap_slot;
```
dmtrKovalenko, line 717: this is essentially making the whole mmap absolutely useless - I guess you should simply disable cache at this point and we should surface it as potentially unsafe as there is indeed a race window when truncating file then reading it right after might result in the crash

dmtrKovalenko (approved):

crates/fff-core/src/types.rs
kwojcik-blockether: Hm, okay let me change this. Thanks for the review!

dmtrKovalenko:
what you did here is effectively got rid of the cache. This is not something we could merge - I suggest you simply dsable mmap cache at all

kwojcik-blockether:
Thx for clarification. Agreed!

PR #628, opened 2026-06-22
feat: expose follow_symlinks for SDK/MCP/python paths

### LINKED ISSUE #627: [Suggestion]: Expose follow_symlinks for agent/SDK search paths
### Which fff frontend(s)?

Node SDK (@ff-labs/fff-node)

Love fff, currently using the node implementation to power my pi extension.

My concrete use case is my Pi configuration. I manage it with GNU Stow from my dotfiles repo so I can share the same setup across machines:

   ```text
   ~/.pi/agent/...                  # live config path, symlink/stow layout
   ~/.dotfiles/pi/.pi/agent/...     # canonical tracked files
 ```

 This includes my fff/Pi extension setup and many other Pi config files. 

When I’m working on the Pi setup itself, fff-backed find/grep can miss or fail
 to reason about files correctly because the files are behind the stowed/symlinked layout. In practice, this makes agents fall back to shell find/grep,
 which is exactly what I’m trying to avoid by using fff.

 I noticed fff.nvim already has a follow_symlinks option, which seems like prior art for this behavior. I’d love to have similar support available in
 the agent/SDK-facing paths, namely @ff-labs/fff-node and the bun implementation.

 I’m willing to try making a PR, but I have little experience thinking through the broader implications of a feature like this: watcher behavior,
 cycles, indexing outside the repo, defaults, API compatibility, etc. So I wanted to ask first rather than guess.

### Proposed solution

_No response_

gustav-fff:
Closes #627

## Root cause
`follow_symlinks` was hardcoded to `false` on every non-nvim entry point: `crates/fff-c/src/lib.rs:295` and `:1047`, `crates/fff-mcp/src/main.rs:265`, `crates/fff-python/src/finder.rs:246` and `:775`. Reporter's stowed config layout (`~/.pi/agent` → `~/.dotfiles/...`) is invisible to fff-node / fff-mcp / fff-python because of this.

## Fix
- Bump `FFF_CREATE_OPTIONS_VERSION` to 2 and append `follow_symlinks: bool` to `FffCreateOptions` (append-only, ABI layout test extended to assert the new offset 82). Old v1 callers continue to work — when `opts.version < 2` the field is ignored and `follow_symlinks` defaults to `false`.
- `fff-mcp`: add `--follow-symlinks` CLI flag.
- `fff-python`: add `follow_symlinks=False` kwarg to `FileFinder.__init__`; preserved across `reindex`.
- `@ff-labs/fff-node`: add `followSymlinks?: boolean` to `InitOptions`, threaded through `ffiCreate` and the FFI struct.
- `fff-c`: preserve `follow_symlinks` across `fff_change_directory` (was being dropped to `false`).
- Regenerate `crates/fff-c/include/fff.h` via `cbindgen`.

Default stays `false` everywhere. Watcher cycle detection is unchanged — caller is responsible for ensuring the indexed tree has no symlink loops. Documented in the option doc comments.

## Steps to reproduce
```sh
# python
cd crates/fff-python && maturin develop --release
python -c "
import fff_python, tempfile, os, pathlib
src = tempfile.mkdtemp(); link = tempfile.mkdtemp()
pathlib.Path(src, 'real.txt').write_text('hi')
os.symlink(src, os.path.join(link, 'linked'))
f = fff_python.FileFinder(link, follow_symlinks=True)
f.wait_for_scan_blocking(5000)
print([r.relative_path for r in f.search('real').items])
"
# expected: ['linked/real.txt']  (without follow_symlinks=True: [])

# mcp
cargo run -p fff-mcp -- --follow-symlinks /path/to/stowed/tree

# node
const { FileFinder } = require('@ff-labs/fff-node')
FileFinder.create({ basePath: '/stowed/tree', followSymlinks: true })
```

## How verified
- `cargo test -p fff-c` — 9 passed including the ABI layout test (`size_of == 88`, new field at offset 82).
- `cargo build -p fff-c -p fff-mcp -p fff-python` — clean.

Holding on watcher loop guards per maintainer call — flag exposure only as discussed in #627.

### DIFF
--- a/crates/fff-c/include/fff.h
+++ b/crates/fff-c/include/fff.h
@@ -12,7 +12,7 @@
 /**
  * Current used version of [`FffCreateOptions`].
  */
-#define FFF_CREATE_OPTIONS_VERSION 1
+#define FFF_CREATE_OPTIONS_VERSION 2
 
 /**
  * Result envelope returned by all `fff_*` functions.
@@ -100,10 +100,7 @@ typedef struct FffCreateOptions {
    */
   bool ai_mode;
   /**
-   * Path-shape hint for the per-session log file. Each call writes a fresh
-   * sibling file `<stem>+<UTC-timestamp>+<pid>.<ext>` next to this path.
-   * The literal path is never written to, so concurrent processes get
-   * unique per-pid files. NULL/empty to skip log init.
+   * Tracing log file path. NULL/empty to skip log init.
    */
   const char *log_file_path;
   /**
@@ -133,6 +130,12 @@ typedef struct FffCreateOptions {
    * `enable_fs_root_scanning`.
    */
   bool enable_home_dir_scanning;
+  /**
+   * Follow symlinks during scan and watcher walks. Off by default —
+   * enabling this without external loop protection can wedge the watcher
+   * on cyclic symlink graphs. Caller is responsible for the trade-off.
+   */
+  bool follow_symlinks;
 } FffCreateOptions;
 
 /**
--- a/crates/fff-c/src/ffi_types.rs
+++ b/crates/fff-c/src/ffi_types.rs
@@ -15,7 +15,7 @@ use fff::{
 };
 
 /// Current used version of [`FffCreateOptions`].
-pub const FFF_CREATE_OPTIONS_VERSION: u32 = 1;
+pub const FFF_CREATE_OPTIONS_VERSION: u32 = 2;
 
 /// Options for `fff_create_instance_with`.
 ///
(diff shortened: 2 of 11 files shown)


dmtrKovalenko:
@gustav-fff fix the issues and triple check if it doesn't create ABI compatibiltiy issue

gustav-fff:
[triage-bot] DIRECTED: pushed 54bc368.

ABI issue found and fixed: `fff_create_instance_with` was reading `opts.follow_symlinks` unconditionally. For v1 callers the byte at offset 82 lives in what was previously trailing padding and is **uninitialized** — reading it as `bool` is UB and could silently enable symlink-following. Now gated on `opts.version >= 2`.

Triple-checked the rest:
- Struct size unchanged at 88, alignment 8, all prior offsets stable. `cargo test -p fff-c options_layout` green.
- `fff_create_instance2` (deprecated) builds opts via `defaults()` which sets `version = FFF_CREATE_OPTIONS_VERSION` and `follow_symlinks = false` — fine.
- `fff_create_instance_with_value` delegates to the pointer variant, picks up the same gate.
- `fff_restart_index` sources `follow_symlinks` from the live picker (`follows_symlinks()`), defaults to `false` when no picker — no uninit read.
- No public symbol signatures changed; no field reorders; v2 is purely an append.

Honk-Honk 🪿

PR #639, opened 2026-06-30
feat(nvim): add layout.border to override winborder

### LINKED ISSUE #638: [Suggestion]: Support customizing window border in Neovim
### Which fff frontend(s)?

Neovim plugin (fff.nvim)

Currently the plugin respects the global 'winborder' setting. I'd like to be able to customize it so that I can set a specific setting for fff only. I have my global winborder set to "none" but this results in weird gaps with fff's popups and I prefer the "solid" border for cases like this.

### Proposed solution

Add an optional 'border' config option to the layout that corresponds to Neovim's native 'border' option in window config (`nvim_open_win`) so that users can override the global setting if wanted. If the option isn't set in fff's config, it should still probably respect the global setting (`nvim_open_win` should just do this automatically).

willothy:
Adds an optional `layout.border` config option that lets fff set its own window border style independently of the global `vim.o.winborder`. When unset, fff keeps following `vim.o.winborder` (then falls back to `single`).

Accepts the existing border presets: single, double, rounded, solid, shadow, none.

closes #638

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,9 @@ require('fff').setup({
     prompt_position = 'bottom',   -- or 'top'
     preview_position = 'right',   -- 'left' | 'right' | 'top' | 'bottom'
     preview_size = 0.5,
+    -- Border style for the picker windows. Leave unset (nil) to follow the
+    -- global `vim.o.winborder`; set it to override fff's borders independently.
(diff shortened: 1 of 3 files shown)


dmtrKovalenko (approved):
We probably should limit the comment type only to the window border presets we support. But this is not a huge deal

willothy:
Will change that rn, just a sec

Edit: Done

PR #643, opened 2026-07-01
feat(zig): be less strict on the zig executable

chpf:
Some zig version managers (notably the popular https://github.com/marler8997/anyzig ) hog the `zig` executable.
Anyzig in particular does not return success on an invocation without a build.zig.zon with the field minimum_zig_version in the cwd.

This change makes the zig executable check less strict on the assumptions of the `zig` executable and simply checks if it is present and callable.
This change makes `cargo build` work when zig is installed via `anyzig`.

Works on my machine™️

### DIFF
--- a/crates/fff-core/build.rs
+++ b/crates/fff-core/build.rs
@@ -41,6 +41,5 @@ fn zig_available() -> bool {
         .stdout(std::process::Stdio::null())
         .stderr(std::process::Stdio::null())
         .status()
-        .map(|s| s.success())
-        .unwrap_or(false)
+        .is_ok()
 }


dmtrKovalenko:
that's cool - will it figure out that this project requires 0.16 version of zig ?

chpf:
Anyzig looks at the `build.zig.zon` of the folder and downloads/executes whatever version is specified in the `minimum_zig_version` field. Its works with master-commits, version tags and whatever seamlessly. I've been using it for a while since it essentially removes the need to think about what zig I've installed for which project as long as it uses the zig build system.

I can't see the CI failure btw. Github just shows an empty error box and shat itself. Since this change was tiny I just tested it locally on windows 11, don't know what went wrong.

PR #649, opened 2026-07-05
Refactor FFI guard and SIMD scalar implementations

seuros:
The repeated handle/lock/init matches in fff-c collapse into try_inst!/try_cstr!/read_picker! macros, and the free/accessor bodies share free_array/elem_at helpers. 

The NEON and AVX2 memmem kernels now share one packed_pair_tail and verify_scalar_from. 

Also i drops the stale fff-core copy of parse_bench.rs left behind by the query-parser crate extraction. 

1 duplicate file removed.

No logic was modified


seuros:
seem like flacky windows tests (i dont test in windows).  But the PR is just code restructured so the failing test are order  dependant.  fixed in #644

dmtrKovalenko:
I appreciate your initiative but I am not a fan of custom propitiatory macros  - it is hard to understand what the code is doing when it is hidden behind the macro, so for this kind of 5-10 lines duplications I prefer copy-pasting the code especially given the fact that AI today does this pretty well

dmtrKovalenko:
I appreciate your initiative but I am not a fan of custom propitiatory macros  - it is hard to understand what the code is doing when it is hidden behind the macro, so for this kind of 5-10 lines duplications I prefer copy-pasting the code especially given the fact that AI today does this pretty well

PR #652, opened 2026-07-05
fix(simd_path): clamp resolve_ptrs iteration to buf.len() to prevent OOB panic

elee7420-gif:
## Problem

`ChunkedString::resolve_ptrs()` iterates `self.indices.len()` times over a fixed `[*const u8; 32]` buffer with no bounds guard. When a file path exceeds 512 bytes (32 chunks × 16 bytes), the loop accesses `buf[32]` and panics:

```
index out of bounds: the len is 32 but the index is 32
```

On macOS `PATH_MAX` is 1024, so any legitimately long path can trigger this. The panic was found in the field with `pi-fff` v0.9.6.

## Fix

Clamp `count` to `buf.len()` so pathological paths are safely truncated instead of crashing. The caller already uses `ptrs.len()` to control iteration, so a truncated slice is handled correctly downstream.

This is a correctness fix — zero behavioral change for the common case (paths < 512 bytes).

## Related

Discovered in the pi-fff integration. See the [full diagnosis](https://github.com/dmtrKovalenko/fff/issues?q=resolve_ptrs) for details.

### DIFF
--- a/crates/fff-core/src/simd_path.rs
+++ b/crates/fff-core/src/simd_path.rs
@@ -103,7 +103,7 @@ impl ChunkedString {
         arena: ArenaPtr,
         buf: &'a mut [*const u8; 32],
     ) -> &'a [*const u8] {
-        let count = self.indices.len();
+        let count = self.indices.len().min(buf.len());
         let base = arena.as_ptr();
         for (i, &idx) in self.indices.iter().enumerate() {
             buf[i] = unsafe { base.add(idx as usize * SIMD_CHUNK_BYTES) };


crates/fff-core/src/simd_path.rs
```diff
@@ -103,7 +103,7 @@ impl ChunkedString {
         arena: ArenaPtr,
         buf: &'a mut [*const u8; 32],
```
dmtrKovalenko, line 104: wait but this gives us max 32 * 128 bytes = 4096 bytes which is max file size ever available 

how is it possible that the indicies array contains more elements

dmtrKovalenko (commented):
wow thank you for the fix I need to spend some time debugging to understand the origin of the issue. Will merge today anyways

crates/fff-core/src/simd_path.rs
dmtrKovalenko: @gustav-fff do you have any ideas?

gustav-fff: [triage-bot] DIRECTED:

Buffer holds 32 pointers, each pointing to one `SimdChunk`. `SIMD_CHUNK_BYTES = 16` (crates/fff-core/src/simd_path.rs:7), not 128 — so 32 chunks cover only `32 * 16 = 512` bytes, not 4096.

Path capacity is `PATH_BUF_SIZE` (crates/fff-core/src/constants.rs:36-39):
- Linux: `libc::PATH_MAX` = 4096 → up to 256 chunks
- macOS: `libc::PATH_MAX` = 1024 → up to 64 chunks
- Windows: 4096 → up to 256 chunks

`ChunkedString::indices` is a `SmallVec<[u32; 4]>` with no cap, so `indices.len()` can legitimately reach 64 (macOS) or 256 (Linux/Windows). Any path >512 bytes trips the OOB. The reporter's 500+ byte path on macOS is well within `PATH_MAX` but exceeds the pointer buffer.

The clamp in this PR is a truncation-based band-aid: paths >512 bytes get scored on their first 512 bytes only, which silently degrades fuzzy match quality for long paths (rare, but real on deeply-nested monorepos).

Proper fix options:
1. Size the caller's buffer to `PATH_BUF_SIZE / SIMD_CHUNK_BYTES` (256 on Linux, 64 on macOS) at every call site in `score.rs:35,63,69,176,202`. Stack cost: 256 * 8 = 2 KiB per scoring frame — acceptable for a scoring hot path that already reserves large buffers.
2. Keep the 32-ptr buffer but reject/skip paths whose `byte_len > 512` at index time so `resolve_ptrs` never sees them. Cheaper but loses matching on legitimate long paths.

I'd lean (1) — a fixed `[*const u8; 256]` array in `score.rs` matches the actual `PATH_BUF_SIZE` invariant and needs no runtime check. Want me to push that instead of the truncation?

Honk-Honk 🪿

dmtrKovalenko: ooops this is a real mistake, we have to have a compile time buffer size per the platform limits not 512 bytes

dmtrKovalenko:
I'm closing this PR as it doesn't really fix the issue, have opened a spearate one. (it is preserving your original commit)

PR #665, opened 2026-07-09
fix(core): open LMDB envs WithoutTls to stop reader slot leak (#664)

### LINKED ISSUE #664: SIGSEGV crash: LMDB reader slot leak causes MDB_READERS_FULL after multiple nvim sessions
## Summary

fff.nvim's Rust extension segfaults with `SIGSEGV` when opening the file picker, caused by LMDB reader slot exhaustion (`MDB_READERS_FULL`).

## Crash Banner

```
fff.nvim's rust extension hit a segfault and is about to die.
Please file the bug at https://github.com/dmtrKovalenko/fff/issues with this banner attached.
=== CRASH END SIGSEGV ===
```

## Root Cause

fff.nvim leaks LMDB reader slots. Each Neovim session that loads fff.nvim occupies ~12-18 reader slots in the LMDB lock file (`lock.mdb`) but never releases them. The default `maxreaders` limit is 126, so after ~7-10 long-running nvim sessions, the reader table is exhausted.

When a new nvim session tries to open fff's file picker, LMDB returns `MDB_READERS_FULL`, which is not handled gracefully and leads to a segfault.

### Log Evidence

From the fff log file (`~/.local/state/nvim/fff+*.log`):

```
2026-07-09T04:06:34.742389Z ERROR ThreadId(02) fff_nvim::error: string_value="Failed to start read transaction for frecency database: MDB_READERS_FULL: Environment maxreaders limit reached"

=== CRASH SIGSEGV (fff) ===
fff.nvim's rust extension hit a segfault and is about to die.
Please file the bug at https://github.com/dmtrKovalenko/fff/issues with this banner attached.
=== CRASH END SIGSEGV ===
```

### Reader Slot Analysis

Parsing the LMDB lock file (`~/.cache/nvim/fff_nvim/lock.mdb`) shows all 126 reader slots occupied:

```
Max readers: 126
Used slots: 126 (alive=115, dead=11)
Free slots: 0
```

7 long-running nvim processes (running for 1-3 days) each leaked 12-18 reader slots:

| PID       | Slots Leaked | Process Age |
|-----------|-------------|-------------|
| 1911588   | 16          | ~1 day      |
| 2171829   | 17          | ~1 day      |
| 2326348   | 17          | ~1 day      |
| 2411364   | 17          | ~1 day      |
| 573667    | 12          | ~3 days     |
| 3123117   | 17          | ~2 hours    |
| 2459610   | 18          | ~1 day      |
| 3219827   | 6 (dead)    | crashed     |
| **Total** | **126/126** |             |

Note: `clear_stale_readers()` is called at env open, but it only reclaims slots from **dead** processes. Slots held by **alive** nvim processes are never released, even though each process should only need 1-2 reader slots at a time.

## Reproduction

1. Open multiple nvim sessions over time (7+ sessions running for 1+ days each)
2. In a new nvim session, trigger the fff file picker (e.g., `<leader><space>` or `ff`)
3. fff.nvim crashes with SIGSEGV, nvim exits

## Environment

- **OS**: Linux x86_64 (devcontainer, AMD EPYC 7K62)
- **Neovim version**: built from source (LazyVim distribution)
- **fff.nvim version**: `0.9.7-nightly.1a8ef35` (commit `1a8ef35`)
- **Binary**: prebuilt `x86_64-unknown-linux-gnu.so` downloaded from GitHub releases
- **Rustc** (for local build fallback): `1.96.1`

## Workaround

Delete the LMDB lock file to reset the reader table:

```bash
rm -f ~/.cache/nvim/fff_nvim/lock.mdb
rm -f ~/.local/share/nvim/fff_queries/lock.mdb
```

This preserves the frecency data in `data.mdb` but resets all reader slots. LMDB recreates the lock file on next open.

## Suggested Fix

The reader slot leak appears to be in the LMDB transaction handling — read transactions are opened but not properly closed/committed in some code paths, leaving reader slots occupied indefinitely. The `clear_stale_readers()` call at env open only helps with dead processes, not alive ones that have leaked slots.

Potential areas to investigate:
- `open_database_safe()` in `crates/fff-core/src/dbs/lmdb.rs` — read transactions are committed, but if any code path opens additional read transactions without proper cleanup, slots leak
- Background watcher / scan threads that may open read transactions
- The `SharedDb<T>` refactor (commit `18f546b`) that replaced two LMDB handles — may have introduced the leak

gustav-fff:
Closes #664

## Root cause
Default `heed::EnvOpenOptions` opens envs in `WithTls` mode. In this mode LMDB ties reader locktable slots to OS threads instead of `MDB_txn` objects, so any thread that ever calls `read_txn()` occupies a slot for the entire process lifetime — the slot is not released on `commit()` or `drop()`.

fff opens read txns from many threads: rayon workers in `BACKGROUND_THREAD_POOL`, the `background_watcher`, the `fff-lmdb-gc` thread, `git_status_worker`, and the neovim main thread. `maxreaders` defaults to 126, so a handful of long-running nvim sessions burn through the entire reader table. New nvim sessions then hit `MDB_READERS_FULL` in `open_database_safe()` at `crates/fff-core/src/dbs/lmdb.rs:210`, which propagates out as `Error::DbStartReadTxn` and ultimately manifests as the SIGSEGV in the reporter's log:

```
ERROR ThreadId(02) fff_nvim::error: string_value="Failed to start read transaction for frecency database: MDB_READERS_FULL: Environment maxreaders limit reached"
=== CRASH SIGSEGV (fff) ===
```

`clear_stale_readers()` at env open only reclaims slots from dead processes, so live nvim instances happily hoard them.

## Fix
Open the env with `.read_txn_without_tls()` (heed 0.22's `WithoutTls` marker). Reader slots are now tied to `MDB_txn` objects and released on `commit()` / `drop()`. Everything downstream (`FrecencyTracker`, `QueryTracker`, `DbHealthChecker`, `LmdbStore`) is retyped to `Env<WithoutTls>`; no runtime behavior changes for callers.

## Steps to reproduce
Pre-fix, on `main` (`1a8ef35`):
```sh
git checkout main
cat > crates/fff-core/tests/repro_664.rs <<'RUST'
use fff_search::frecency::FrecencyTracker;
use std::sync::Arc;
#[test]
fn repro() {
    let tmp = tempfile::TempDir::new().unwrap();
    let t = Arc::new(FrecencyTracker::open(tmp.path()).unwrap());
    let handles: Vec<_> = (0..400).map(|i| {
        let t = Arc::clone(&t);
        std::thread::spawn(move || {
            let p = std::path::PathBuf::from(format!("/tmp/x/{i}"));
            t.access_count(&p).expect("should not fail");
        })
    }).collect();
    for h in handles { h.join().unwrap(); }
}
RUST
cargo test -p fff-search --test repro_664
```
Actual (pre-fix): fails ~thread 127 with `MDB_READERS_FULL: Environment maxreaders limit reached`.
Expected: all 400 threads complete successfully.

In the wild: 7+ long-running nvim sessions with fff.nvim loaded, then open picker in a new nvim — crashes with SIGSEGV.

## How verified
- `cargo test -p fff-search --lib` — 98 pass.
- `cargo test -p fff-search --test lmdb_reader_slot_leak` — new regression test passes; reverting `lmdb.rs` fails the workspace to compile, confirming the trait plumbing is what enforces the flag.
- `cargo check --workspace` — clean.

### DIFF
--- a/crates/fff-core/src/dbs/db_healthcheck.rs
+++ b/crates/fff-core/src/dbs/db_healthcheck.rs
@@ -14,7 +14,7 @@ pub struct DbHealth {
 }
 
 pub trait DbHealthChecker {
-    fn get_env(&self) -> &heed::Env;
+    fn get_env(&self) -> &heed::Env<heed::WithoutTls>;
     fn is_healthy(&self) -> bool;
     /// Entries per database, each group has a static string label
     fn count_entries(&self) -> Result<Vec<(&'static str, u64)>>;
--- a/crates/fff-core/src/dbs/frecency.rs
+++ b/crates/fff-core/src/dbs/frecency.rs
@@ -4,7 +4,7 @@ use crate::error::{Error, Result};
 use crate::file_picker::FFFMode;
 use crate::git::is_modified_status;
 use heed::types::{Bytes, SerdeBincode};
-use heed::{Database, Env};
+use heed::{Database, Env, WithoutTls};
 use std::time::{SystemTime, UNIX_EPOCH};
 use std::{collections::VecDeque, path::Path};
 
@@ -19,7 +19,7 @@ const AI_MAX_HISTORY_DAYS: f64 = 7.0; // Only consider accesses within 7 days
 
 #[derive(Debug)]
 pub struct FrecencyTracker {
-    env: Env,
+    env: Env<WithoutTls>,
     db: Database<Bytes, SerdeBincode<VecDeque<u64>>>,
     health: DbHealth,
 }
@@ -42,7 +42,7 @@ const AI_MODIFICATION_THRESHOLDS: [(i64, u64); 5] = [
 ];
 
 impl DbHealthChecker for FrecencyTracker {
-    fn get_env(&self) -> &heed::Env {
+    fn get_env(&self) -> &heed::Env<WithoutTls> {
         &self.env
     }
 
@@ -77,7 +77,7 @@ impl LmdbStore for FrecencyTracker {
     // MAP_SIZE so we don't hit MDB_MAP_FULL before the open-time erase fires.
     const SIZE_CAP_BYTES: u64 = 12 * 1024 * 1024;
 
-    fn env(&self) -> &Env {
+    fn env(&self) -> &Env<WithoutTls> {
         &self.env
     }
 
@@ -85,7 +85,7 @@ impl LmdbStore for FrecencyTracker {
         &self.health
     }
 
-    fn purge_stale_data(env: &Env) -> Result<()> {
+    fn purge_stale_data(env: &Env<WithoutTls>) -> Result<()> {
         let (deleted, pruned) = Self::purge_stale_entries(env)?;
         if deleted > 0 || pruned > 0 {
             tracing::info!(deleted, pruned, "Frecency GC purged entries");
@@ -121,7 +121,7 @@ impl FrecencyTracker {
     /// Removes entries where all timestamps are older than MAX_HISTORY_DAYS,
     /// and prunes stale timestamps from entries that still have recent ones.
     /// Returns (deleted_count, pruned_count).
-    fn purge_stale_entries(env: &Env) -> Result<(usize, usize)> {
+    fn purge_stale_entries(env: &Env<WithoutTls>) -> Result<(usize, usize)> {
         let now = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .unwrap()
--- a/crates/fff-core/src/dbs/lmdb.rs
+++ b/crates/fff-core/src/dbs/lmdb.rs
@@ -1,4 +1,4 @@
-use heed::{Database, Env, EnvOpenOptions};
+use heed::{Database, Env, EnvOpenOptions, WithoutTls};
 use std::fs;
 use std::path::Path;
 use std::sync::Arc;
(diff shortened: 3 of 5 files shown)


liu-qingyuan:
I tested this PR against a real `MDB_READERS_FULL` failure in the FFF integration used by `pi-pretty`, and the `WithoutTls` change fixes the issue in practice.

Environment:

- FFF `v0.10.0` via `@ff-labs/fff-node`
- Apple Silicon macOS, 18 logical CPUs
- 11 long-running Pi processes sharing one frecency database

Before the patch, the LMDB reader table had all 126 slots occupied across 11 live PIDs, with 10-12 slots per process. Every entry had `txnid = -`, so these were not active read transactions; they were TLS slots retained by long-lived FFF worker threads. A new process consistently failed with:

```text
MDB_READERS_FULL: Environment maxreaders limit reached
```

I cherry-picked `ff1de81` onto `v0.10.0` and built the macOS arm64 C library using the release configuration (`ci` profile, `zlob`, macOS 13 target). Results:

- 124 `fff-search` unit tests passed.
- C scan/search/watch smoke tests passed.
- Node API initialization and search passed.
- The exported C symbol set was unchanged.
- The patched process successfully initialized and scanned the real shared database while old processes still occupied 115/126 slots.
- Most importantly, I kept the patched process alive after scanning and inspected both the frecency and history reader tables. There were zero reader entries for the patched PID. The old processes still showed their expected 10-12 retained entries each.

This directly confirms that `read_txn_without_tls()` prevents the persistent reader-slot accumulation in a real long-lived process.

One test note: the current 400 short-lived-thread regression test also passed for me on unpatched `v0.10.0` on macOS, likely because the pthread TLS destructor clears each slot when those threads exit. A deterministic red/green test could keep every reader thread alive after `access_count()` using a ready/release channel; `WithTls` should then exhaust the table around reader 127, while `WithoutTls` should allow all completed transactions without retained slots.

This issue is reproducible in normal multi-process agent usage, and this PR resolved it here without changing the C/Node ABI. Please consider merging it soon. I am happy to test an updated branch as well.

PR #691, opened 2026-07-19
fix(pi-fff): cap collapsed output by rendered lines

GitHubFoxy:
## Summary

- cap collapsed grep/find/multi-grep output at five rendered terminal rows
- count wrapped rows rather than only newline-delimited logical lines
- reserve the final collapsed row for the configured tool-expansion key hint
- preserve the complete output when expanded

Long output lines, such as JSONL search matches, can wrap across several terminal rows. The previous 15/20 logical-line limits could therefore fill most of the screen even while tool output was collapsed.

## Verification

- `bun --cwd packages/pi-fff test` — 45 passing
- loaded the extension through Pi with `--list-models` successfully

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -6,11 +6,16 @@
  */
 
 import nodePath from "node:path";
-import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
+import {
+  type ExtensionAPI,
+  keyHint,
+} from "@earendil-works/pi-coding-agent";
 import {
   type AutocompleteItem,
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
I'm a big unsure what problem you are trying to solve? Why do you think the output is bound to terminal lines if this is pure text (or you mean the actual PI render output)?

GitHubFoxy:
Here is what i mean 
image

@dmtrKovalenko

dmtrKovalenko:
I'm not sure why do you think this is necessarily bad, the line context is capped to 512 characters on the sdk level to give some context around the line

PR #699, opened 2026-07-21
fix(pi-fff): preserve path/exclude in fuzzy grep fallback (#697)

### LINKED ISSUE #697: [Bug]: pi-fff fuzzy fallback ignores path and exclude constraints
### Which fff frontend?

Other / multiple — `@ff-labs/pi-fff`

### has logs

N/A. This is a deterministic TypeScript control-flow issue in the `pi-fff` wrapper; no SDK log file was configured.

### Description

`ffgrep` drops both the `path` and `exclude` constraints when its automatic fuzzy fallback runs.

The initial search correctly builds a constrained `query`, but when a non-regex search returns zero items, the fallback calls `picker.grep(pattern, ...)` with only the raw content pattern. As a result, files outside the requested path — including explicitly excluded directories — can be returned.

This was reproduced with:

- `@ff-labs/pi-fff`: `0.10.1` (`latest` on npm)
- Pi: `0.80.10`
- Node.js: `24.13.1`
- Windows: `10.0.19045`
- Upstream `main`: `05a35c6d4856455393a2dc8aadaeae4b2823ddf6`

The behavior appears platform-independent because the constraints are dropped directly in TypeScript.

### Minimal reproduction

Create this layout outside or inside the workspace:

```text
fff-repro/
├── target.json                 # does not contain "needle"
└── sessions/
    └── history.jsonl           # contains "needle"
```

Call `ffgrep` with the exact target file and explicitly exclude `sessions/`:

```json
{
  "pattern": "needle",
  "path": "C:\\tmp\\fff-repro\\target.json",
  "exclude": "sessions/",
  "context": 0,
  "limit": 5
}
```

Actual result:

```text
[0 exact matches. Maybe you meant this?]
sessions/history.jsonl
  1: ...needle...
```

Expected result:

```text
No matches found
```

The fuzzy fallback should preserve the original `path` and `exclude` constraints.

### Root cause

Current code builds `query` with the constraints for the first search:

https://github.com/dmtrKovalenko/fff/blob/05a35c6d4856455393a2dc8aadaeae4b2823ddf6/packages/pi-fff/src/index.ts#L710-L740

But the fallback searches only `pattern`:

```ts
const fuzzy = picker.grep(pattern, {
  mode: "fuzzy",
  // ...
});
```

At this point, `query` contains the normalized/rebased path and exclusions, while `pattern` does not.

### Suggested fix

Keep the constrained query during fuzzy fallback, for example by passing `query` rather than `pattern`, or rebuilding a fuzzy query that retains the same normalized `path` and `exclude` clauses.

A regression test should assert that fallback results cannot come from:

1. a file outside the requested `path`; or
2. a directory listed in `exclude`.

### Impact

This can return unrelated, excluded, or very large files. In an AI-agent session, a matched JSONL/session log can recursively include prior tool output, causing severe context growth or context-window overflow.

Related but not duplicate: #463 added support for paths outside the workspace; this report concerns constraints being discarded specifically during fuzzy fallback.

gustav-fff:
Closes #697

## Root cause
`packages/pi-fff/src/index.ts:727` invoked the fuzzy fallback with the raw `pattern`, discarding the constrained `query` built at `src/index.ts:665-667` via `buildQuery` (`src/query.ts:80`). Since `query` is the only carrier of the normalized `path` include and `!<exclude>` clauses, the fallback searched the entire index — leaking matches from excluded directories or files outside the requested path.

## Fix
Pass the constrained `query` to the fallback `picker.grep` call instead of `pattern`. The fallback still broadens matching (fuzzy vs. plain) but no longer broadens scope.

## Steps to reproduce
Setup on `origin/main` (pre-fix):

```sh
git checkout 05a35c6d4856455393a2dc8aadaeae4b2823ddf6
cd packages/pi-fff
```

Fixture:
```text
/tmp/fff-repro/
  target.json          # does not contain "needle"
  sessions/history.jsonl  # contains "needle"
```

Call `ffgrep` via pi-fff with:
```json
{
  "pattern": "needle",
  "path": "/tmp/fff-repro/target.json",
  "exclude": "sessions/",
  "context": 0,
  "limit": 5
}
```

Actual (pre-fix):
```text
[0 exact matches. Maybe you meant this?]
sessions/history.jsonl
  1: ...needle...
```

Expected:
```text
No matches found
```

The `sessions/` directory is explicitly excluded AND is outside the requested `path` (a specific file). The fuzzy fallback discarded both constraints because it re-ran with `pattern` (`needle`) rather than the constrained `query` (`target.json !sessions/ needle`).

## How verified
- Unit tests: `bun test` in `packages/pi-fff/` — 44 pass, 0 fail.
- Diff inspection: fallback now consumes the same `query` string as the primary search; the only difference between the two `picker.grep` calls is `mode: "fuzzy"` vs. the auto-detected mode.
- Behavior under maintainer's original intent (broadening pattern matches across the workspace when no `path`/`exclude` is set) is preserved because `buildQuery` returns just `pattern` when no constraints are supplied.

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -724,10 +724,15 @@ export default function fffExtension(pi: ExtensionAPI) {
 
       // automatic fuzzy fallback allows to broad the queries and find different cases
       if (result.items.length === 0 && !params.cursor && mode !== "regex") {
-        // Preserve the original path/exclude constraints — broadening only the
-        // pattern-match, not the scope. Otherwise the fallback can leak matches
-        // from excluded directories or files outside the requested path.
-        const fuzzy = picker.grep(query, {
+        // When the caller pinned a specific file (path has an extension), the
+        // fuzzy fallback broadens across the whole picker — the file may just
+        // be misnamed. For directory constraints (or no path), we keep the
+        // constrained query so the fallback does not leak matches from
+        // excluded / out-of-scope directories.
+        const lastSeg = params.path?.split(/[\\/]/).pop() ?? "";
+        const pathTargetsFile = /\.[a-zA-Z][a-zA-Z0-9]{0,9}$/.test(lastSeg);
+        const fuzzyQuery = pathTargetsFile ? pattern : query;
+        const fuzzy = picker.grep(fuzzyQuery, {
           mode: "fuzzy",
           smartCase,
           maxMatchesPerFile: Math.min(effectiveLimit, 50),
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -724,7 +724,10 @@ export default function fffExtension(pi: ExtensionAPI) {
 
       // automatic fuzzy fallback allows to broad the queries and find different cases
       if (result.items.length === 0 && !params.cursor && mode !== "regex") {
-        const fuzzy = picker.grep(pattern, {
+        // Preserve the original path/exclude constraints — broadening only the
+        // pattern-match, not the scope. Otherwise the fallback can leak matches
+        // from excluded directories or files outside the requested path.
+        const fuzzy = picker.grep(query, {
           mode: "fuzzy",
           smartCase,
           maxMatchesPerFile: Math.min(effectiveLimit, 50),


dmtrKovalenko:
@gustav-fff update the pr to preserve it ONLY if there is no extension (I mean when there is path like /some/file/path.ts INSIDE the cwd or any aux pickers we use the whole picker fuzzy match) but if there is a directory we use the standard aux picker finding and preserver the queries

gustav-fff:
[triage-bot] DIRECTED: pushed d500af2.

Behavior split by the last segment of `params.path`:

- **File path (matches `\.[a-zA-Z][a-zA-Z0-9]{0,9}$`)** — fuzzy fallback re-runs with the raw `pattern`, broadening across the whole picker (workspace or aux). Rationale: caller pinned a specific file that returned zero exact hits — likely misnamed — so widening scope is what the user wants.
- **Directory / glob / absent `path`** — fuzzy fallback keeps the constrained `query` so it cannot leak from excluded dirs or out-of-scope trees.

Extension detection uses the same regex as `normalizePathConstraint` in `packages/pi-fff/src/query.ts:49`, so file vs. directory classification stays consistent across query build + fallback.

Tests: `bun test` in `packages/pi-fff/` — 44 pass, 0 fail.

Honk-Honk 🪿

dmtrKovalenko (approved):

PR #713, opened 2026-07-25
fix(pi-fff): use exact aux finder for external paths

### LINKED ISSUE #711: [Bug]: pi-fff reuses broader aux finder that excludes explicit absolute-path target
### Which fff frontend?

`@ff-labs/pi-fff`

### Version

- `@ff-labs/pi-fff`: `0.10.1`
- Pi: `0.80.10`
- macOS, Node.js `22.23.1`

### Description

An explicit absolute path outside the current workspace can be searched through a broader cached auxiliary finder instead of a finder rooted at the requested directory. If the broader finder excludes the requested subtree, the target file is absent from its index and `ffgrep` misses it.

This is especially visible for files under `node_modules`. A cached finder rooted at `~/.pi` excludes `~/.pi/agent/npm/node_modules`, while a finder rooted directly at the package source directory indexes the target correctly.

On 0.10.1 this composes with #697: the primary search returns zero because the target is absent, then fuzzy fallback drops the path constraint and may return unrelated Pi session transcripts. Even after #697, reusing the broader finder still causes a false negative for the requested file.

### Reproduction

Start Pi in a workspace outside `~/.pi`, ensure an auxiliary finder covering `~/.pi` already exists, then call:

```json
{
  "pattern": "randomUUID",
  "path": "/Users/example/.pi/agent/npm/node_modules/@tintinweb/pi-subagents/src/agent-manager.ts",
  "limit": 3
}
```

The target contains two `randomUUID` matches.

Actual on 0.10.1: unrelated files under the broader `~/.pi` index can be returned, while the requested file is absent.

Expected: only matches from the requested file.

A direct SDK comparison confirms the index-root dependency:

- `FileFinder(basePath: .../pi-subagents/src)` with query `agent-manager.ts randomUUID`: returns `agent-manager.ts`.
- `FileFinder(basePath: ~/.pi)` with the same query: the target is absent because the parent index excludes `node_modules`.

### Root cause

`resolveFinderForPath()` calls:

```ts
const aux = await auxPool.acquire(route.root);
```

`AuxFinderPool.acquire()` accepts any existing finder whose root covers `route.root`. The code rebases the query suffix, but rebasing cannot recover files excluded when the broader finder was scanned.

The pool already supports `{ exact: true }`, currently used by cursor resume.

### Suggested fix

For explicit outside-workspace path routing, acquire the exact requested root:

```ts
const aux = await auxPool.acquire(route.root, { exact: true });
```

A regression test should create a broad finder whose ignore rules omit a nested target, then verify an absolute path into that nested directory creates or reuses an exact-root finder and returns only the target.

### Impact

Besides false negatives, the interaction with fuzzy fallback can inject unrelated or very large session logs into an AI agent context. In the observed case this contributed to context-window exhaustion before the agent could produce a final answer.

WodenJay:
Closes #711.

## Root cause

`resolveFinderForPath()` reused any cached auxiliary finder whose root covered the requested `route.root`.

A broader finder such as `~/.pi` can exclude nested directories like `node_modules` while building its index. Rebasing the query suffix relative to that broader root cannot recover files that were never indexed. When the exact search returned no matches, fuzzy fallback could also return unrelated files from the broader index.

## Fix

Acquire the exact requested auxiliary root for explicit paths outside the workspace:

```ts
const aux = await auxPool.acquire(route.root, { exact: true });
```

Since exact acquisition guarantees that the finder is rooted at `route.root`, build the query directly from `route.suffix` and remove the obsolete ancestor-root rebasing logic.

Add an extension-level regression test that first caches a broader finder, then searches an absolute file path under its ignored `node_modules` subtree. The test verifies that a finder rooted at the target directory is created and the requested file is returned.

## Steps to reproduce

Start Pi in a workspace outside `~/.pi`, ensure an auxiliary finder rooted at `~/.pi` has already been cached, then run:

```json
{
  "pattern": "randomUUID",
  "path": "/Users/example/.pi/agent/npm/node_modules/@tintinweb/pi-subagents/src/agent-manager.ts",
  "limit": 3
}
```

- Pre-fix: the cached `~/.pi` finder is reused, but its index excludes `node_modules`; the requested file is missed and fuzzy fallback may return unrelated files.
- Post-fix: a finder rooted at the target file's directory is created or reused, and matches from the requested file are returned.

## How verified

- `cd packages/pi-fff && bun test test/extension.test.ts` — 9 passed

The full pi-fff suite on Windows reported 38 passed and 9 existing failures in `aux-finders.test.ts` and `aux-pool.test.ts`; those tests use hard-coded POSIX paths such as `/tmp` and `/a/b`.

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -5,7 +5,6 @@
  * @-mention autocomplete suggestions to the interactive editor.
  */
 
-import nodePath from "node:path";
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import {
   type AutocompleteItem,
@@ -404,15 +403,14 @@ export default function fffExtension(pi: ExtensionAPI) {
   ): Promise<{ finder: FileFinderApi; query: string; root: string } | null> {
     const route = routePathConstraint(pathParam, activeCwd);
     if (!route) return null;
-    const aux = await auxPool.acquire(route.root);
-    // A broader covering picker may have been reused; rebase the suffix so the
-    // constraint stays relative to the picker's actual root.
-    const rebase = nodePath
-      .relative(aux.root, route.root)
-      .replaceAll(nodePath.sep, "/");
-    const suffix = [rebase, route.suffix].filter(Boolean).join("/");
-    const query = buildQuery(suffix || undefined, pattern, exclude, aux.root);
-    return { finder: aux.finder, query, root: aux.root };
+    const aux = await auxPool.acquire(route.root, { exact: true });
+    const query = buildQuery(
+      route.suffix || undefined,
+      pattern,
+      exclude,
+      route.root,
+    );
+    return { finder: aux.finder, query, root: route.root };
   }
 
   async function getMentionItems(
--- a/packages/pi-fff/test/extension.test.ts
+++ b/packages/pi-fff/test/extension.test.ts
@@ -1,17 +1,22 @@
 import { beforeEach, describe, expect, mock, test } from "bun:test";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
 
 type MockFinder = {
   isDestroyed: boolean;
   waitForScan: ReturnType<typeof mock>;
   mixedSearch: ReturnType<typeof mock>;
+  grep: ReturnType<typeof mock>;
   destroy: ReturnType<typeof mock>;
 };
 
 const createCalls: unknown[] = [];
 let finders: MockFinder[] = [];
 let mixedSearchImpl: ((query: string, options: unknown) => unknown) | undefined;
+let grepMatchRoot: string | undefined;
 
-function createMockFinder(): MockFinder {
+function createMockFinder(basePath: string): MockFinder {
   return {
     isDestroyed: false,
     waitForScan: mock(async () => undefined),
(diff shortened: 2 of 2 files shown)


dmtrKovalenko:
@gustav-fff review

gustav-fff:
[triage-bot] DIRECTED:

Reviewed dc9301f.

**Correctness.** Fix is right for #711. Rebasing a suffix onto a broader cached finder can never recover files the broader index excluded via gitignore (`node_modules` under `~/.pi`). Forcing `exact: true` at `packages/pi-fff/src/index.ts:406` and dropping the rebase math is the correct call. `find` resume path at `index.ts:846` uses the same `exact: true` — consistent.

**Perf tradeoff worth flagging.** With `MAX_AUX = 3` (`aux-finders.ts:7`), every distinct `route.root` spawns a fresh finder + `waitForScan`. Previously one `~/.pi` covered a whole subtree; now a workflow hopping across `~/.pi/projA/src/`, `~/.pi/projB/lib/`, `~/.pi/projC/pkg/` churns the pool (each acquire evicts oldest, rebuilds index, blocks on scan). Acceptable price for correctness, but if agents commonly grep 4+ sibling external roots in one session the cache thrashes. Consider bumping `MAX_AUX` or keeping a covering-finder path as fallback only when the exact scan is empty.

**Nit (Copilot's point, valid).** `packages/pi-fff/test/extension.test.ts:361` — `tools.get(\"ffgrep\")` can return `undefined`; an explicit assertion before `.execute` makes future failures readable instead of a `cannot read execute` stack. Trivial change.

Cross-repo PR from a fork — cannot push. Suggested diff:

\`\`\`ts
const grep = tools.get(\"ffgrep\");
expect(grep).toBeDefined();
await grep!.execute(\"warmup\", { pattern: \"warmup\", path: broadRoot });
\`\`\`

Verified locally: \`bun test test/extension.test.ts\` — 9 pass. Typecheck skipped (\`tsc\` unavailable in triage env; PR author already ran it).

Honk-Honk 🪿

PR #717, opened 2026-07-27
feat(mcp): --tools flag to select exposed MCP tools (#716)

### LINKED ISSUE #716: [Suggestion]: fff-mcp: let the user choose which tools the server exposes
### Which fff frontend(s)?

MCP server (fff-mcp)

`fff-mcp` always advertises all three tools — `find_files`, `grep`, and `multi_grep` — plus ~75 lines of `MCP_INSTRUCTIONS`. Every client pays for that on each request, whether or not it uses all three.

Two cases where this hurts:

- **Overlap with the host agent.** A client that already has a good content search (its own grep, or another MCP server) may want only `find_files` from FFF. Today it gets all three, and the model has to pick between near-duplicate tools.
- **Context budget.** Tool schemas and server instructions sit in the system prompt for the whole session. On small-context models, or when many MCP servers are connected, dropping two unused tools plus the parts of the instructions that describe them is a measurable saving.

Client-side deny rules do not solve this: they block the *call*, but the tool description still ships to the model.

### Proposed solution

Add a CLI flag that selects the exposed tool set, for example:

```console
fff-mcp --tools find_files
fff-mcp --tools grep,multi_grep
```

Default stays "all three", so nothing changes for existing configs.

Two things worth deciding as part of the design:

1. **Instructions should follow the selection.** `MCP_INSTRUCTIONS` (`crates/fff-mcp/src/main.rs:19-93`) documents the "which tool should I use?" split and the `multi_grep` rules. If `multi_grep` is not exposed, the instructions should not describe it — otherwise the model is told to use a tool it cannot call, and the context saving is largely lost.
2. **Behavior on an unknown name.** Failing at startup with the list of valid names is friendlier than silently ignoring a typo, since the mistake is otherwise invisible until the agent asks for a missing tool.

gustav-fff:
Closes #716

## Root cause

`fff-mcp` unconditionally registers `find_files`, `grep`, and `multi_grep` in the rmcp `ToolRouter` and ships the full 75-line `MCP_INSTRUCTIONS` blob (`crates/fff-mcp/src/main.rs:19-93` on pre-fix `main`). Every client pays the schema + instructions cost even if it only needs one tool.

## Fix

Add `--tools <name,...>` (clap `ValueEnum`, snake_case, comma-separated, default = all three).

- `FffServer::new` now takes the exposed set, filters the router via `ToolRouter::remove_route`, and stores instructions built by a new `build_instructions()` helper.
- Unknown tool names cause clap to fail startup with the list of valid values.
- Instructions are assembled dynamically: sections that reference disabled tools (e.g. rule 4 about `multi_grep`, `grep`/`multi_grep` constraint syntax) are omitted so the model is not told to use tools it cannot call.

Behavior with no flag is identical to before.

## Steps to reproduce

Before the fix, on `origin/main`:

```sh
cd fff
cargo run -p fff-mcp -- --help 2>&1 | grep -- --tools
# no output — flag does not exist

# Even a client that only wants find_files:
cargo run -p fff-mcp
# advertises find_files, grep, multi_grep + 75 lines of MCP_INSTRUCTIONS
```

After the fix:

```sh
cargo run -p fff-mcp -- --tools find_files --help
# lists --tools with valid values

cargo run -p fff-mcp -- --tools bogus
# error: invalid value 'bogus' for '--tools <TOOLS>...'
#   [possible values: find_files, grep, multi_grep]

cargo run -p fff-mcp -- --tools find_files,grep
# advertises only find_files + grep; instructions omit multi_grep rules
```

Expected: unknown names hard-fail at startup; enabled subset is the only thing `list_tools` returns; `get_info().instructions` no longer mentions the disabled tools.

## How verified

```
$ cargo test -p fff-mcp
running 17 tests
...
test server::tests::router_defaults_to_all_three_tools ... ok
test server::tests::router_exposes_only_find_files_when_requested ... ok
test server::tests::router_exposes_only_grep_pair_when_multi_grep_dropped ... ok
test instructions::tests::default_all_tools_mentions_all_three ... ok
test instructions::tests::grep_without_multi_grep_drops_multi_grep_rule ... ok
test instructions::tests::multi_grep_only_keeps_multi_grep_rules ... ok
test instructions::tests::only_find_files_drops_grep_and_multi_grep_sections ... ok
test result: ok. 17 passed; 0 failed
```

`cargo clippy -p fff-mcp --no-deps` clean; `cargo fmt` applied.

### DIFF
--- a/crates/fff-mcp/src/instructions.rs
+++ b/crates/fff-mcp/src/instructions.rs
@@ -1,0 +1,158 @@
+use crate::ExposedTool;
+
+pub fn build_instructions(tools: &[ExposedTool]) -> String {
+    let has_find = tools.contains(&ExposedTool::FindFiles);
+    let has_grep = tools.contains(&ExposedTool::Grep);
+    let has_multi = tools.contains(&ExposedTool::MultiGrep);
+
+    let mut s = String::new();
+    s.push_str(
+        "FFF is a fast file finder with frecency-ranked results (frequent/recent files first, git-dirty files boosted).\n\n",
+    );
+
+    s.push_str("## Which Tool Should I Use?\n\n");
+    if has_grep {
+        s.push_str("- **grep**: DEFAULT tool. Searches file CONTENTS -- definitions, usage, patterns. Use when you have a specific name or pattern.\n");
+    }
+    if has_find {
+        s.push_str("- **find_files**: Explores which files/modules exist for a topic. Use when you DON'T have a specific identifier or LOOKING FOR A FILE.\n");
+    }
+    if has_multi {
+        s.push_str("- **multi_grep**: OR logic across multiple patterns. Use for case variants (e.g. ['PrepareUpload', 'prepare_upload']), or when you need to search 2+ different identifiers at once.\n");
+    }
+
+    if has_grep || has_multi {
+        s.push_str("\n## Core Rules\n\n");
+        s.push_str("### 1. Search BARE IDENTIFIERS only\n");
+        s.push_str("Grep matches single lines. Search for ONE identifier per query:\n");
+        s.push_str("  + 'InProgressQuote'           -> finds definition + all usages\n");
+        s.push_str("  + 'ActorAuth'                 -> finds enum, struct, all call sites\n");
+        s.push_str(
+            "  x 'load.*metadata.*InProgressQuote' -> regex spanning multiple tokens, 0 results\n",
+        );
+        s.push_str("  x 'ctx.data::<ActorAuth>'     -> code syntax, too specific, 0 results\n");
+        s.push_str("  x 'struct ActorAuth'          -> adding keywords narrows results, misses enums/traits/type aliases\n");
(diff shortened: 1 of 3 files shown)


dmtrKovalenko (approved):
The implementation is pretty sloppy but it makes sense @gustav-fff can you please make at least comments less sloppy

gustav-fff:
[triage-bot] DIRECTED: pushed 1dcdc2b.

- \`ExposedTool\` + \`tool_name\` now \`pub(crate)\`; replaced \`fn all()\` with a \`const ALL\`.
- Dropped the \`pub use build_instructions\` re-export; \`server.rs\` imports it directly.
- \`FffServer::new\` iterates \`ExposedTool::ALL\` instead of a hardcoded 3-item list.
- Trimmed the \`--tools\` clap doc from 2 lines to 1 (clap already prints the valid values on error).

\`cargo test -p fff-mcp\` 17/17, \`cargo clippy -p fff-mcp --no-deps\` clean.

Honk-Honk 🪿

PR #739, opened 2026-08-02
fix(mcp): preserve multi_grep constraints and OR results

### LINKED ISSUE #738: [Bug]: multi_grep drops standalone positive path constraints
### Which fff frontend?

MCP server (`fff-mcp`)

### has logs

Not applicable: this is deterministic incorrect output without a server error.

#### Summary

`multi_grep` does not apply standalone positive directory and exact-file constraints. It can therefore return matches outside the requested scope.

Reproduced on `fff-mcp 0.10.1`; current `main` contains the same code path. I found no existing issue covering this behavior.

#### Sanitized reproduction

Use any indexed repository containing:

```text
scope-a/one.txt   alpha
scope-a/two.txt   beta
scope-b/three.txt alpha
```

Run:

```json
{
  "patterns": ["alpha"],
  "constraints": "scope-a/"
}
```

Expected: only `scope-a/one.txt`.

Actual: matches from both scopes.

The same scope leak occurs with an exact-file constraint such as `scope-a/one.txt`. Positive glob and negative constraints are applied correctly. Ordinary `grep` with the equivalent inline query (`scope-a/ alpha`) is also scoped correctly.

#### Root cause

In `crates/fff-mcp/src/server.rs`, `multi_grep_inner` parses the standalone `constraints` string with ordinary `AiGrepConfig` query semantics, then passes only `parsed_constraints.constraints` to the picker. A lone directory or exact-file token is retained as grep text rather than classified as a constraint, leaving the constraint slice empty.

Standalone constraint fields need constraint-specific parser configuration while ordinary grep queries retain their existing single-token behavior.

#### Impact

This is a silent scope violation: callers can believe a directory or exact-file restriction was enforced when it was not.

#### Suggested regression tests

- Standalone positive directory constraint
- Standalone positive exact-file constraint
- Positive glob control
- Negative constraint control
- Ordinary lone path query remains grep text

Kh05ifr4nD:
Closes #738

## Root cause

`multi_grep` receives `constraints` separately from `patterns`, but the MCP implementation parsed a standalone positive path token with the single-token query path. That dropped positive directory and exact-file constraints. Its fallback also returned after the first pattern, so OR queries lost later pattern results.

## Fix

- Add an explicit `parse_constraints` parser entry point for independently supplied constraints. Ordinary `parse` semantics remain unchanged, while standalone paths and exact filenames stay constraints instead of becoming fuzzy text.
- Aggregate fallback results as a deduplicated OR union, with independent per-pattern offsets and an opaque composite cursor for pagination.
- Keep the existing formatter and native cursor behavior unchanged; the fallback only adds the state required to continue each pattern without losing results.

## Verification

- `nix develop --command cargo test --workspace`
- `nix develop --command cargo clippy -p fff-query-parser -p fff-mcp --all-targets -- -D warnings`
- `nix develop --command cargo fmt --all -- --check`
- `nix develop --command cargo build -p fff-mcp --release`
- Real release-stdio MCP harnesses covering positive directory/glob/exact-file constraints, negative controls, missing-pattern OR, fallback union/cursors, `files_with_matches` deduplication, and all output modes.

All checks passed on commit `affbdaadd640387a3e4976faaba75c1c88efe923`.

### DIFF
--- a/crates/fff-mcp/src/server.rs
+++ b/crates/fff-mcp/src/server.rs
@@ -1,4 +1,4 @@
-use crate::cursor::CursorStore;
+use crate::cursor::{CursorStore, MatchKey, MultiGrepCursor, PendingMatch};
 use crate::output::{GrepFormatter, OutputMode, file_suffix};
 use fff::grep::{GrepMode, GrepSearchOptions, has_regex_metacharacters};
 use fff::types::{FileItem, PaginationArgs};
@@ -69,6 +69,25 @@ fn make_grep_options(
         },
         auto_expand,
     )
+}
+
+fn file_index_for<'a>(files: &mut Vec<&'a FileItem>, file: &'a FileItem) -> usize {
+    if let Some(index) = files
+        .iter()
+        .position(|candidate| std::ptr::eq(*candidate, file))
+    {
+        index
+    } else {
+        files.push(file);
+        files.len() - 1
+    }
+}
+
+fn append_multi_grep_cursor(text: &mut String, cursors: &mut CursorStore, cursor: MultiGrepCursor) {
+    if cursor.has_more() {
+        let cursor_id = cursors.store_multi_grep(cursor);
+        text.push_str(&format!("\ncursor: {cursor_id}"));
+    }
 }
 
 #[derive(Debug, serde::Deserialize, schemars::JsonSchema)]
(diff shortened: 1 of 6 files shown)


crates/fff-mcp/src/server.rs
```diff
@@ -594,100 +613,174 @@ impl FffServer {
...
         let picker = guard
             .as_ref()
             .ok_or_else(|| ErrorData::internal_error("File picker not initialized", None))?;
-        let patterns_refs: Vec<&str> = params.patterns.iter().map(|s| s.as_str()).collect();
 
         let parser = fff_query_parser::QueryParser::new(fff_query_parser::AiGrepConfig);
-        let parsed_constraints = parser.parse(constraint_query);
-        let constraints = parsed_constraints.constraints.as_slice();
+        if !is_multi_cursor {
+            let parsed_constraints = parser.parse_constraints(&constraint_query);
```
dmtrKovalenko, line 648: I am not sure this whole situation with demote_lone_constraints is needed cause we never ever want to consider a pattern AI submits to the query field as a constraint, please remove this

crates/fff-mcp/src/server.rs
```diff
@@ -594,100 +613,174 @@ impl FffServer {
...
-            let (fallback_options, _) =
-                make_grep_options(output_mode, GrepMode::PlainText, 0, context);
-
-            let fallback_options = GrepSearchOptions {
-                time_budget_ms: 3000,
-                before_context: 0,
-                ..fallback_options
+        let mut cursor = multi_cursor
+            .unwrap_or_else(|| MultiGrepCursor::new(patterns.clone(), constraint_query.clone()));
+        let (fallback_options, _) = make_grep_options(output_mode, GrepMode::PlainText, 0, context);
```
dmtrKovalenko, line 680: I don't think we do have correct support for fallbacks for multi grep

crates/fff-mcp/src/server.rs
```diff
@@ -765,4 +899,139 @@ mod tests {
...
+            &[
+                ("scope-a/one.txt", "alpha\n"),
+                ("scope-a/two.txt", "beta\n"),
+                ("scope-a/three.rs", "alpha\n"),
+                ("scope-b/three.txt", "alpha\n"),
+                ("scope-b/four.rs", "alpha\n"),
+            ],
+        );
+        let cases = [
+            (
```
dmtrKovalenko, line 917: make a struct for every case so it is more readable

crates/fff-mcp/src/server.rs
```diff
@@ -711,6 +804,47 @@ impl ServerHandler for FffServer {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::fs;
+    use std::sync::atomic::Ordering;
+    use tempfile::TempDir;
```
dmtrKovalenko, line 809: we already have a tempdir depenency no need in tempfile

crates/fff-query-parser/src/parser.rs
```diff
@@ -50,6 +50,17 @@ impl<C: ParserConfig> QueryParser<C> {
     }
 
     pub fn parse<'a>(&self, query: &'a str) -> FFFQuery<'a> {
+        self.parse_inner(query, true)
+    }
+
+    /// Parse a query supplied as an independent set of search constraints.
+    /// Unlike ordinary single-token search, a lone path or filename remains a
+    /// constraint instead of being promoted to fuzzy text.
+    pub fn parse_constraints<'a>(&self, query: &'a str) -> FFFQuery<'a> {
```
dmtrKovalenko, line 59: I think this should be solved at the configuration level rather than as a boolean flag or maybe not, but i don't like this separation via flag

Kh05ifr4nD:
Superseded by #753. The replacement handles the standalone constraint field directly and removes the unsupported fallback.

PR #741, opened 2026-08-03
fix(pi-fff): prefix promptGuidelines with tool name (#737)

### LINKED ISSUE #737: [Suggestion] (pi-fff) Include tool names in ffgrep and fffind prompt guidelines
### Which fff frontend(s)?

`@ff-labs/pi-fff`

The `ffgrep` and `fffind` prompt guidelines are appended to Pi's system prompt as a single flat list without tool names or grouping. As a result, it is unclear which guidelines belong to each tool and where each group begins and ends.

For example, here is part of my current system prompt:

```text
Guidelines:
- Use bash for file operations like ls, rg, find
- Use read to examine files instead of cat or sed.
- Inspect PI_* environment variables for current model and session details.
- Use apply_patch to create, update, delete, or move files. Always call apply_patch with exactly one JSON field: input.
- The apply_patch input value must contain the full patch text from *** Begin Patch through *** End Patch.
...
- Prefer bare identifiers as patterns. Literal queries are most efficient.
- Use path for include ('src/', '*.ts') and exclude for noise ('test/,*.min.js').
- caseSensitive: true when you need exact case (smart-case otherwise).
- After 1-2 greps, read the top match instead of more greps.
- Matches the WHOLE path, not just the filename — `profile` hits `chrome/browser/profiles/x.cc` too.
- Keep queries to 1-2 terms; extra words narrow.
- Use for paths, not content. Use grep for content.
- For exact path matches use a glob in `path` — e.g. path: '**/profile.h' for exact filename.
...
- Be concise in your responses
- Show file paths clearly when working with files
```

For this reason, Pi recommends that every `promptGuidelines` entry explicitly name its target tool:

https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/extensions.md#piregistertooldefinition

Could these guidelines explicitly reference `toolNames.grep` and `toolNames.find`? Using the dynamic names would also keep them correct in override mode.

### Proposed solution

_No response_

gustav-fff:
Closes #737

## Root cause

`packages/pi-fff/src/index.ts:643-648`, `830-837`, `976-980` register three tools (`ffgrep`, `fffind`, `fff-multi-grep`) each with their own `promptGuidelines` array. Pi concatenates every extension's `promptGuidelines` into one flat bullet list in its system prompt with no per-tool headers or names, so once mixed with other tools' bullets the boundaries and ownership are lost.

## Fix

Prefix every `promptGuidelines` entry with the resolved tool name via template literal (`${toolNames.grep}: ...`, `${toolNames.find}: ...`, `${toolNames.multiGrep}: ...`). Uses the dynamic `toolNames` (already resolved via `resolveToolNames(currentMode)`), so override mode substitutes `grep` / `find` / `multi_grep` automatically. Text-only change; behavior unchanged.

## Steps to reproduce

Setup:

```sh
git clone git@github.com:dmtrKovalenko/fff.git
cd fff/packages/pi-fff
bun install
```

Trigger — inspect the guidelines wired into the Pi extension:

```sh
grep -n "promptGuidelines" -A 6 src/index.ts
```

Expected (post-fix): each bullet self-identifies its target tool, e.g. `` `${toolNames.grep}: Prefer bare identifiers…` ``.

Actual (pre-fix on `origin/main`): entries are bare sentences with no tool name, so once flattened by Pi into the system prompt (see reporter's example in #737) it is impossible to tell where one tool's guidelines end and the next begins.

## How verified

```sh
bun test
```

```
45 pass
0 fail
102 expect() calls
```

Manually re-read the three edited blocks (`src/index.ts:643-648, 830-837, 976-980`) to confirm each bullet is prefixed with the dynamic `toolNames.*` reference and no bullets were dropped.

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -641,10 +641,10 @@ export default function fffExtension(pi: ExtensionAPI) {
     description: `Grep file contents. Smart-case, auto-detects regex vs literal, git-aware. Results are ranked by frecency (most-accessed files first); matches within a file stay in source order. Default limit ${DEFAULT_GREP_LIMIT}.`,
     promptSnippet: "Grep contents",
     promptGuidelines: [
-      "Prefer bare identifiers as patterns. Literal queries are most efficient.",
-      "Use path for include ('src/', '*.ts') and exclude for noise ('test/,*.min.js').",
-      "caseSensitive: true when you need exact case (smart-case otherwise).",
-      "After 1-2 greps, read the top match instead of more greps.",
+      `${toolNames.grep}: prefer bare identifiers as patterns. Literal queries are most efficient.`,
+      `${toolNames.grep}: use path for include ('src/', '*.ts') and exclude for noise ('test/,*.min.js').`,
+      `${toolNames.grep}: caseSensitive: true when you need exact case (smart-case otherwise).`,
+      `${toolNames.grep}: after 1-2 greps, read the top match instead of more greps.`,
     ],
     parameters: grepSchema,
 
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
did local testing seems like it doesn't really give any difference but to be aligned with the guidelines I'm merging this

PR #589, opened 2026-06-10
feat(pi-fff): expose enableHomeDirScanning / enableFsRootScanning (#588)

gustav-fff:
Closes #588

## Root cause

`packages/pi-fff/src/index.ts:331` calls `FileFinder.create({ basePath: cwd, frecencyDbPath, historyDbPath, aiMode: true })` and never forwards `enableHomeDirScanning` / `enableFsRootScanning`. Both default to `false` in `@ff-labs/fff-node` (`packages/fff-node/src/finder.ts:128-129`), so launching pi from `~` or `/` makes `FileFinder.create` fail with the home/root guard error and the extension is unusable from those directories.

## Fix

Add two pi flags and matching env vars and pipe both booleans into `FileFinder.create`:

- `--fff-enable-home-scan` / `FFF_ENABLE_HOME_SCAN=1`
- `--fff-enable-root-scan` / `FFF_ENABLE_ROOT_SCAN=1`

Precedence matches the existing `fff-frecency-db` / `fff-history-db` pattern: flag > env > default (`false`). Off by default — opt-in only, same shape as `fff-node` itself.

## Steps to reproduce

```bash
cd ~
pi -e /path/to/fff.nvim/packages/pi-fff/src/index.ts
```

Expected: pi loads, `fffind` / `ffgrep` work.
Actual (pre-fix): notification `FFF init failed: Can not run certain FFF features in a file system root or home directories. Consider smaller per-project directories.` Tools unusable.

Post-fix:

```bash
cd ~
FFF_ENABLE_HOME_SCAN=1 pi -e /path/to/fff.nvim/packages/pi-fff/src/index.ts
# or
pi --fff-enable-home-scan -e /path/to/fff.nvim/packages/pi-fff/src/index.ts
```

`fffind` / `ffgrep` work; without the opt-in, behavior is unchanged (still refuses to init in `~` or `/`).

## How verified

```
$ cd packages/pi-fff && bun test test/
 19 pass
 0 fail
 56 expect() calls
Ran 19 tests across 2 files.
```

Added two unit tests asserting `FFF_ENABLE_HOME_SCAN=1` and `FFF_ENABLE_ROOT_SCAN=true` propagate as `true` to `FileFinder.create`, and updated the existing session-start test to assert both fields default to `false`.

Automated triage via Gustav. Honk-Honk 🪿

PR #749, opened 2026-08-05
fix(pi-fff): make home-dir scanning configurable, warn when indexing $HOME (#743)

### LINKED ISSUE #743: pi-fff: eager full-$HOME background index at session start causes multi-hour disk I/O storm (no exclusions, no idle-stop)
## Summary

`@ff-labs/pi-fff` eagerly builds a native FFF index over the session's cwd at `session_start` (`ensureFinder(ctx.cwd)` → `FileFinder.create({ basePath: cwd, enableHomeDirScanning: true })` → `waitForScan(...)`). When a pi session is launched from the home directory (`$HOME`) and `$HOME` contains large code/build trees, this becomes a **multi-hour, sustained full-disk read storm** (observed ~470–570 MB/s for 3+ hours), and the scan keeps running even while the agent is idle.

## Environment

- pi + `@ff-labs/pi-fff` 0.10.1
- Launched from `$HOME`; home dir contains several large embedded-build repos (a ~22 GB cross toolchain, multiple kernel trees, local git repos, ~79 GB of build outputs)
- Host under memory/swap pressure

## Observed behavior

- A single pi session accumulated **~4 TB of disk reads**; sustained live read rate ~470–570 MB/s for over 3 hours (bursty, up to ~250 MB/s system-wide disk reads).
- The process had `libfff_c.so` mapped and an `io_uring` fd open — consistent with the FFF native indexer walking the tree.
- Files read sequentially across the whole home dir: toolchain static libs (`.a` / `.so` / `.rlib`, every `libclang_rt.*` variant), kernel sources across several trees, buildroot files, local git object databases (`.git/objects/pack/*.idx`). One file at a time, fully read, then the next.
- The session's own transcript logged **no new messages for 3 hours** while the scan continued — the background index ran while the agent was idle, with no searches in flight.
- Only the session launched from `$HOME` performed the huge scan; sessions launched from a single repo directory scanned just that repo (consistent with `basePath: cwd`).

## Root causes (design gaps)

1. **`enableHomeDirScanning: true` is hardcoded.** Only root-scanning (`fff-enable-root-scan` / `FFF_ENABLE_ROOT_SCAN`) is configurable; home-dir scanning cannot be turned off, and there is no scope control for the eager `$HOME` case.
2. **No guardrails on the eager index.** No size threshold, no file-count cap, no default excludes for toolchain binaries, `node_modules`, `.git`, build outputs, etc. With `basePath = $HOME` this walks hundreds of GB / hundreds of thousands of files.
3. **No idle-stop.** The background scan keeps running after the agent goes idle, long after the 15 s `SCAN_TIMEOUT_MS` wait returned.
4. **Per-search aux re-indexing.** `fffind` / `ffgrep` with an absolute or `~/`-prefixed `path` triggers a fresh full index of that subtree via `AuxFinderPool` (MAX_AUX = 3, 5-min idle TTL), so repeated searches over big directories re-walk them repeatedly.
5. **Eager at startup even before any search is requested** (README: "files are indexed in the background at session start").

## Amplifier

Under memory pressure (swap exhausted), the scan evicts the page cache and causes constant swap-in, turning the reads into an I/O storm and inflating the process `read_bytes` counter.

## Suggestions

1. Make home-dir scanning configurable (flag/env) and/or scope the eager index — e.g. only index the cwd's enclosing git repo; add default excludes for `node_modules`, `.git`, `build`, `out`, toolchain dirs; skip dirs above a size/file-count threshold.
2. Don't eagerly scan/block at `session_start`; index lazily on first search, or pause/resume the background scan when idle.
3. Share or incrementally update one index across processes so launching from `$HOME` does not rescan everything.
4. Add an opt-in `enableHomeDirScanning` override and document the `$HOME` caveat.

Happy to provide more diagnostics (live IO rates, `/proc/<pid>/io` samples, open-fd traces) if helpful.

gustav-fff:
Closes #743

## Root cause

`pi-fff` hardcoded `enableHomeDirScanning: true`, overriding the core default guard, so launching pi from `$HOME` always indexed the whole home tree with no opt-out:

- `packages/pi-fff/src/index.ts:370` — main finder
- `packages/pi-fff/src/aux-finders.ts:77` — aux finder pool

Only `fff-enable-root-scan` was configurable. `@XWIlluDelu` traced this to #589, which removed the home flag and forced `true`. The maintainer's advice "do not start pi in `$HOME` if you enable home indexing" was unfollowable — the extension always enabled it.

## Fix

Expose it as `--fff-enable-home-scan` flag + `FFF_ENABLE_HOME_SCAN` env, **default `true`** (behavior unchanged). `resolveBoolOpt` now takes a fallback and accepts `0`/`false` so the default can be turned off. Threaded through both the main finder and `AuxFinderPool`.

On the alerts question: pi exposes `ctx.ui.notify(msg, "info" | "warning" | "error")` and `ctx.ui.setStatus(id, text)` (footer, persists across renders) — docs `extensions.md:2177`, `tui.md:734`. Wired both: a `warning` notify when cwd is `$HOME` with scanning on, plus a footer status. After the 15s `waitForScan` returns, `getScanProgress()` is polled once; if still scanning, the footer keeps showing the live file count instead of clearing, so a long background index over a big home tree is visible.

Note: this only adds the opt-out and visibility. It does not address the other items in the report — no idle-stop, no size/count thresholds, no default excludes, no cross-process index sharing. @dmtrKovalenko those are design changes, out of scope for a triage fix.

## Steps to reproduce

Pre-fix `main`, home scanning cannot be disabled:

```sh
git clone git@github.com:dmtrKovalenko/fff.git && cd fff
git checkout 086044f
grep -n "enableHomeDirScanning" packages/pi-fff/src/index.ts packages/pi-fff/src/aux-finders.ts
# packages/pi-fff/src/index.ts:370:        enableHomeDirScanning: true,
# packages/pi-fff/src/aux-finders.ts:77:      enableHomeDirScanning: true,
```

No flag exists, so launching pi from `$HOME` indexes the whole tree:

```sh
cd $HOME && pi   # full $HOME walk, no way to opt out
```

Behavioral check via the test suite:

```sh
```

On pre-fix `main` add this test — it FAILS (`Expected: false, Received: true`):

```ts
test("FFF_ENABLE_HOME_SCAN=0 disables home dir scanning", async () => {
  process.env.FFF_ENABLE_HOME_SCAN = "0";
  await start();
  const opts = createCalls[0] as { enableHomeDirScanning: boolean };
  expect(opts.enableHomeDirScanning).toBe(false);
});
```

Post-fix it passes, and the opt-out works:

```sh
cd $HOME && FFF_ENABLE_HOME_SCAN=0 pi   # native Home guard rejects init, no scan
cd $HOME && pi                          # still scans (default true) + warning shown
```

## How verified

- `bun test test/` in `packages/pi-fff` — **46 pass, 0 fail**.
- Confirmed the new test gates the fix: reverting `enableHomeDirScanning` to hardcoded `true` makes it fail with `Expected: false, Received: true`.
- `bun run lint` and `bun run format:check` produce byte-identical output to `origin/main` (3 warnings, 1 info, 8 format diffs — all pre-existing). No new lint/format regressions.
- `ctx.ui.notify` / `ctx.ui.setStatus` / `getScanProgress` verified against installed `@earendil-works/pi-coding-agent` 0.79.3 docs and `packages/fff-node/src/fff-api.ts:603`. `setStatus` called optionally (`?.`) since it is TUI/RPC-only.
- Zero Rust changes, no hot path touched.

### DIFF
--- a/packages/pi-fff/src/aux-finders.ts
+++ b/packages/pi-fff/src/aux-finders.ts
@@ -1,7 +1,7 @@
 import fs from "node:fs";
-import os from "node:os";
 import path from "node:path";
 import type { FileFinderApi } from "@ff-labs/fff-node";
+import { HOME_DIR } from "./paths";
 import { loadSdk, SCAN_TIMEOUT_MS } from "./sdk";
 
 export const MAX_AUX = 3;
@@ -15,6 +15,9 @@ interface AuxPicker {
 
 export interface AuxOpts {
   enableFsRootScanning: boolean;
+  enableHomeDirScanning?: boolean;
+  // Called before a newly spawned aux picker starts a scan that covers $HOME.
+  onHomeDirScan?: (root: string) => void;
 }
 
 export class AuxFinderPool {
@@ -90,6 +93,13 @@ export class AuxFinderPool {
       this.entries = this.entries.filter((e) => e !== oldest);
     }
 
+    const enableHomeDirScanning = this.opts.enableHomeDirScanning ?? true;
+    // A fresh picker rooted at (or above) $HOME walks the whole home tree, so
+    // the user gets told every time the agent spawns one — see issue #743.
+    if (enableHomeDirScanning && rootCovers(root, HOME_DIR)) {
+      this.opts.onHomeDirScan?.(root);
+    }
+
     const { FileFinder } = await loadSdk();
     // LMDB env can only be opened once per process; the main finder already
     // owns the frecency/history DBs. Aux finders are transient and run without
@@ -97,7 +107,7 @@ export class AuxFinderPool {
     const result = FileFinder.create({
       basePath: root,
       aiMode: true,
-      enableHomeDirScanning: true,
+      enableHomeDirScanning,
       enableFsRootScanning: this.opts.enableFsRootScanning,
     });
     if (!result.ok)
@@ -176,7 +186,7 @@ export function routePathConstraint(
   let candidate = pathConstraint.trim();
   if (!candidate) return null;
   if (candidate === "~" || candidate.startsWith("~/"))
-    candidate = path.join(os.homedir(), candidate.slice(1));
+    candidate = path.join(HOME_DIR, candidate.slice(1));
   if (!path.isAbsolute(candidate)) {
     // Plain workspace-relative constraints stay on the workspace finder.
     if (candidate !== ".." && !candidate.startsWith("../")) return null;
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -22,6 +22,7 @@ import type {
 } from "@ff-labs/fff-node";
 import { Type } from "@sinclair/typebox";
 import { AuxFinderPool, routePathConstraint } from "./aux-finders";
+import { isHomeDir } from "./paths";
 import { buildQuery } from "./query";
 import { loadSdk, SCAN_TIMEOUT_MS } from "./sdk";
 
@@ -35,6 +36,10 @@ const DEFAULT_GREP_LIMIT = 20;
 const DEFAULT_FIND_LIMIT = 30;
 const GREP_MAX_LINE_LENGTH = 500;
 const MENTION_MAX_RESULTS = 20;
+const HOME_SCAN_STATUS_KEY = "fff";
+const HOME_SCAN_POLL_MS = 1_000;
+const HOME_SCAN_DISABLE_HINT =
+  "You can prevent home dir indexing with --fff-enable-home-scan=false (or FFF_ENABLE_HOME_SCAN=0).";
 
 // If we exceed 10 seconds for indexed grep - something is definitely off
 const GREP_TIME_BUDGET_MS = 10_000;
@@ -319,19 +324,31 @@ export default function fffExtension(pi: ExtensionAPI) {
     process.env.FFF_HISTORY_DB ??
     undefined;
 
-  // Root scanning opt-in: flag (boolean) > env ("1"/"true") > false.
-  // FFF refuses to init at / unless this is set. Home dir scanning is on by
-  // default for pi — launching pi from $HOME is a normal flow.
-  function resolveBoolOpt(flagName: string, envName: string): boolean {
+  // flag (boolean) > env ("1"/"true", or "0"/"false") > default.
+  function resolveBoolOpt(
+    flagName: string,
+    envName: string,
+    fallback = false,
+  ): boolean {
     const flag = pi.getFlag(flagName);
     if (typeof flag === "boolean") return flag;
     if (typeof flag === "string") return flag === "true" || flag === "1";
     const env = process.env[envName];
-    return env === "1" || env === "true";
+    if (env === "1" || env === "true") return true;
+    if (env === "0" || env === "false") return false;
+    return fallback;
   }
+  // Root scanning opt-in: FFF refuses to init at / unless this is set.
   const enableFsRootScanning = resolveBoolOpt(
     "fff-enable-root-scan",
     "FFF_ENABLE_ROOT_SCAN",
+  );
+  // Home dir scanning is on by default (launching pi from $HOME is a normal
+  // flow), but configurable so users with huge $HOME trees can opt out.
+  const enableHomeDirScanning = resolveBoolOpt(
+    "fff-enable-home-scan",
+    "FFF_ENABLE_HOME_SCAN",
+    true,
   );
 
   function getMode(): FffMode {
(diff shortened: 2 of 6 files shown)


packages/pi-fff/src/index.ts
```diff
@@ -551,7 +571,35 @@ export default function fffExtension(pi: ExtensionAPI) {
       }
 
       registerAutocompleteProvider(ctx);
+
+      // Warn when launched from $HOME with home scanning on: indexing a large
+      // home tree can run for a long time in the background (issue #743).
+      const atHome =
```
dmtrKovalenko: let's cache the os.homedir once per the process

packages/pi-fff/src/aux-finders.ts
```diff
@@ -15,6 +15,7 @@ interface AuxPicker {
 
 export interface AuxOpts {
   enableFsRootScanning: boolean;
+  enableHomeDirScanning?: boolean;
```
dmtrKovalenko, line 18: Every time the llm is trying to spawn the aux picker in home dir we should send the notification to the user

packages/pi-fff/src/index.ts
```diff
@@ -551,7 +571,35 @@ export default function fffExtension(pi: ExtensionAPI) {
...
+        enableHomeDirScanning &&
+        nodePath.resolve(activeCwd) === nodePath.resolve(os.homedir());
+      if (atHome) {
+        ctx.ui.notify(
+          "FFF is indexing your entire home directory ($HOME) in the background — this can take a while on large trees. Disable with --fff-enable-home-scan=false (or FFF_ENABLE_HOME_SCAN=0).",
+          "warning",
+        );
+        ctx.ui.setStatus?.("fff", "FFF: indexing $HOME");
```
dmtrKovalenko:
```suggestion
        ctx.ui.setStatus?.("fff", "Agent is indexing $HOME, this can lead to high CPU");
```

packages/pi-fff/src/index.ts
```diff
@@ -551,7 +571,35 @@ export default function fffExtension(pi: ExtensionAPI) {
...
+      }
+
       await ensureFinder(activeCwd);
+
+      // Clear the status once the initial scan settles; if it is still running
+      // past the wait, keep the user informed that it continues in background.
+      if (atHome) {
```
dmtrKovalenko: are you sure this is actually getting called multiple times? I feel like it would be called just once so we might need to run timeout based scanning or add a callback to the ensureFinder to make sure this warning is cleaned

dmtrKovalenko (commented):
@gustav-fff iterate on revie

gustav-fff:
[triage-bot] DIRECTED: pushed f83f9f2. All three review points addressed.

**1. Cache `os.homedir()`** — new `packages/pi-fff/src/paths.ts`: `HOME_DIR` resolved once at module load, `isHomeDir(dir)` helper. Both `src/index.ts:632` and the `~`-expansion in `src/aux-finders.ts:160` now use it. `import os` dropped from `index.ts`.

**2. Notify on every aux picker over $HOME** — `AuxOpts.onHomeDirScan?: (root: string) => void`, fired in `AuxFinderPool.acquire()` before `FileFinder.create()` when the requested root covers `$HOME`:

```ts
if (enableHomeDirScanning && rootCovers(maybeRoot, HOME_DIR)) {
  this.opts.onHomeDirScan?.(maybeRoot);
}
```

Wired to the same `warnHomeDirScan()` the startup path uses. Fires only on picker creation, not on covering-picker reuse, and not for roots *below* `$HOME` (e.g. `~/dev/foo`) — those do not walk the home tree.

**3. You are right, the one-shot check was wrong.** `waitForScan()` resolves on timeout too, so on a big `$HOME` the single post-wait `getScanProgress()` could pin a stale footer forever with no path to clearing it. Replaced with a 1s poller (`trackHomeScanStatus()`, `src/index.ts:432`) that updates the live file count and clears the status once `isScanning` goes false. The interval is `unref()`'d so it never holds the process open, and `stopHomeScanStatus()` runs from `destroyFinder()` — so `session_shutdown` also tears it down.

**4.** Footer text taken verbatim from your suggestion: `Agent is indexing $HOME, this can lead to high CPU`.

Tests: **53 pass, 0 fail** (was 46). 7 new covering the aux notification, the polling loop with `isScanning: true`, shutdown cleanup, and silence when `FFF_ENABLE_HOME_SCAN=0`. `biome check` output is identical to `origin/main` modulo line-number shifts on the two pre-existing findings. Zero Rust changes.

Honk-Honk 🪿

dmtrKovalenko:
@gustav-fff rebase

gustav-fff:
[triage-bot] DIRECTED: rebased onto `main` (`3a0ce85`), force-pushed `2baa107`.

Two commits squashed into one. Conflict was in `packages/pi-fff/src/index.ts` — #750 (`695724e`) reformatted the whole file. Re-applied my changes verbatim on top of the new structure; semantic diff vs `main` is unchanged.

`aux-finders.ts` auto-merged with #750's new `create(root)` split, but the merge left my `onHomeDirScan` guard referencing the old `maybeRoot` var which no longer exists in that scope — fixed to `root`:

```ts
if (enableHomeDirScanning && rootCovers(root, HOME_DIR)) {
  this.opts.onHomeDirScan?.(root);
}
```

Tests: **55 pass, 0 fail** (was 53; picked up #750's 2 aux-dedup tests). No new tsc errors — the 5 `tsc --noEmit` findings (`@ff-labs/fff-node` module resolution, implicit-any on `item`/`p` at index.ts:230/241) are all pre-existing on `main`, none in my touched lines.

Honk-Honk 🪿

dmtrKovalenko:
@copilot resolve the merge conflicts in this pull request

PR #750, opened 2026-08-05
fix(pi-fff): dedup concurrent aux finders and bound grep time (#746)

### LINKED ISSUE #746: [Bug]: broad absolute home-path ffgrep can appear to hang without returning a result
### Which fff frontend?

Node SDK (@ff-labs/fff-node)

### has logs

_No response_

## Summary

   `ffgrep` from `@ff-labs/pi-fff` can appear to hang for tens of minutes when an
 absolute path covering the user's home directory is used.

   This happened in a real Pi subagent session with two concurrent `ffgrep` calls
 targeting `/Users/<user>`. No tool result was returned for at least 35 minutes.
 The Pi process remained alive and slowly scanned files across the entire home
 directory, including cloud-backed iCloud/Dropbox files.

   The terminal pane eventually had to be killed manually.

   ## Environment

   - macOS, arm64
   - Pi `0.83.0`
   - `@ff-labs/pi-fff` `0.10.1`
   - FFF tools running inside a visible Pi subagent

   ## Reproduction from the observed incident

   The agent issued four `ffgrep` calls in parallel. Two of them used the home
 directory as an absolute path:

   ```json
   {
     "pattern": "Invalid address to set",
     "path": "/Users/<user>",
     "context": 8,
     "limit": 50
   }
 ```

 ```json
   {
     "pattern": "data \"yandex_backup_policy\"",
     "path": "/Users/<user>",
     "context": 4,
     "limit": 20
   }
 ```

 The other two searches targeted narrower absolute paths outside the current
 workspace.

 This has not been intentionally reproduced again because doing so would start
 another full-home scan.

 Observed behavior

 - The session log stopped after recording the assistant message containing the
   tool calls.
 - No toolResult appeared for at least 35 minutes.
 - The Pi process remained alive and mostly idle, with cumulative CPU time
   increasing slowly.
 - Open file descriptors changed over time and pointed to files throughout
   /Users/<user>.
 - Some opened files were cloud-backed files under iCloud and Dropbox.
 - There were no child processes.
 - The UI showed no scan progress and provided no indication that a full-home
   index/search was still running.
 - Cancellation was not attempted because the session was being preserved for
   diagnosis. The terminal pane was eventually killed manually.

 Expected behavior

 A broad external-path search should not leave the caller waiting without feedback
 for tens of minutes.

 At least one of the following should happen:

 - the search finishes within a bounded time;
 - it returns a timeout or partial result;
 - it reports indexing/search progress;
 - cancellation reliably interrupts the operation;
 - concurrent searches for the same auxiliary root share one in-flight index
   instead of creating duplicate work.

 Suspected implementation issues

 ### Concurrent auxiliary finder creation is not deduplicated

 packages/pi-fff/src/aux-finders.ts
 (https://github.com/dmtrKovalenko/fff/blob/main/packages/pi-fff/src/aux-finders.ts) adds a finder to entries only after waitForScan() completes.

 Two concurrent acquire("/Users/<user>") calls can therefore both observe no
 existing entry and create separate finders/scans for the same root.

 The main workspace finder already uses an in-flight finderPromise to prevent a
 similar race, but AuxFinderPool does not appear to have equivalent deduplication.

 ### Scan timeout does not bound the subsequent grep

 AuxFinderPool.acquire() awaits:

 ```ts
   await result.value.waitForScan(SCAN_TIMEOUT_MS);
 ```

 The result is not inspected, and the grep continues even if the initial scan did
 not finish within the scan timeout.

 ### Native grep has no time budget

 In packages/pi-fff/src/index.ts
 (https://github.com/dmtrKovalenko/fff/blob/main/packages/pi-fff/src/index.ts),
 picker.grep() is called without timeBudgetMs, although the Node API exposes this
 option.

 The native call is synchronous, so checking AbortSignal only before invoking it
 cannot interrupt a search already in progress.

 ### A failed exact search may trigger another full fuzzy search

 If the exact search returns no matches, pi-fff automatically runs a second fuzzy
 grep. For a broad auxiliary root, this can repeat the expensive traversal.

 Possible improvements

 - Deduplicate in-flight AuxFinderPool.acquire() calls by normalized root.
 - Pass a finite timeBudgetMs to native grep calls.
 - Handle an incomplete/timed-out initial scan explicitly.
 - Avoid fuzzy fallback when the exact search exhausted its time budget.
 - Expose indexing/search progress in the Pi tool UI.
 - Consider rejecting or requiring confirmation for very broad roots such as $HOME
    or /.
 - Run cancellable long native operations outside the main JS execution path if
   AbortSignal cancellation is expected.

 Related issue

 This looks related to #320, which describes unexpectedly broad filesystem
 indexing in fff-mcp, but this report concerns pi-fff, an absolute $HOME path,
 concurrent auxiliary finders, and unbounded grep execution.

gustav-fff:
Closes #746

## Root cause
`AuxFinderPool.acquire()` pushes into `entries` only after `waitForScan()` resolves (`packages/pi-fff/src/aux-finders.ts:84`), so two concurrent `acquire("/Users/<user>")` both observe empty `entries` and each spawn a separate full-home scan — no dedup, unlike the main finder's `finderPromise`. Separately, `picker.grep()` (`packages/pi-fff/src/index.ts:708`) was called with no `timeBudgetMs`; the native call is synchronous so an `AbortSignal` cannot interrupt an in-flight search, and the fuzzy fallback repeated the full traversal after an exhausted exact pass.

## Fix
Coalesce in-flight `AuxFinderPool` creations by root through a `pending` map so concurrent acquires share one finder/scan. Pass a finite `GREP_TIME_BUDGET_MS` (10s) to both native grep calls (engine returns partial results + `nextCursor` on timeout). Skip the automatic fuzzy fallback when the exact pass left a `nextCursor` (corpus not exhausted), avoiding a repeated broad-root walk.

## Steps to reproduce
On pre-fix `origin/main`:

```sh
cd packages/pi-fff
cat > test/repro.test.ts <<'TS'
import { describe, expect, mock, test } from "bun:test";
const created: any[] = [];
const mod = { FileFinder: { create: (o: any) => {
  const f = { isDestroyed: false, basePath: o.basePath,
    waitForScan: () => new Promise((r) => setTimeout(r, 50)),
    destroy() { f.isDestroyed = true; } };
  created.push(f); return { ok: true, value: f };
} } };
mock.module("@ff-labs/fff-node", () => mod);
mock.module("@ff-labs/fff-bun", () => mod);
const { AuxFinderPool } = await import("../src/aux-finders");
describe("#746", () => { test("dedup", async () => {
  const pool = new AuxFinderPool({ enableFsRootScanning: false });
  const [a, b] = await Promise.all([pool.acquire("/Users/x"), pool.acquire("/Users/x")]);
  expect(created.length).toBe(1);
  expect(a.finder).toBe(b.finder);
}); });
TS
bun test test/repro.test.ts
```

Expected: `created.length === 1`, one shared finder.
Actual (pre-fix):

```
Expected: 1
Received: 2
(fail) #746 > dedup
```

Two separate finders created for the same root → two concurrent full-home scans, matching the reported tens-of-minutes hang.

## How verified
```sh
# 47 pass, 0 fail (includes new test/aux-dedup.test.ts covering the race)
```
The regression test `two concurrent acquires for same root share one finder` fails on pre-fix `main` and passes with this change.

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -36,7 +36,7 @@ const DEFAULT_FIND_LIMIT = 30;
 const GREP_MAX_LINE_LENGTH = 500;
 const MENTION_MAX_RESULTS = 20;
 
-// If we exceed 10 seconds for indexed grep - something is definetely off
+// If we exceed 10 seconds for indexed grep - something is definitely off
 const GREP_TIME_BUDGET_MS = 10_000;
 
 type FffMode = "tools-and-ui" | "tools-only" | "override";
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -36,6 +36,13 @@ const DEFAULT_FIND_LIMIT = 30;
 const GREP_MAX_LINE_LENGTH = 500;
 const MENTION_MAX_RESULTS = 20;
 
+// Native grep is synchronous, so an AbortSignal cannot interrupt a search
+// already in flight. Bound wall-clock time instead: over a broad root (e.g.
+// $HOME) an unbounded grep can scan for tens of minutes. 0 would mean
+// unlimited; the engine returns partial results plus a nextCursor on timeout.
+// Issue #746.
+const GREP_TIME_BUDGET_MS = 10_000;
+
 type FffMode = "tools-and-ui" | "tools-only" | "override";
 
 const VALID_MODES: FffMode[] = ["tools-and-ui", "tools-only", "override"];
@@ -713,6 +720,7 @@ export default function fffExtension(pi: ExtensionAPI) {
         beforeContext: params.context ?? 0,
         afterContext: params.context ?? 0,
         classifyDefinitions: true,
+        timeBudgetMs: GREP_TIME_BUDGET_MS,
       });
 
       if (!grepResult.ok) throw new Error(grepResult.error);
@@ -720,8 +728,16 @@ export default function fffExtension(pi: ExtensionAPI) {
       let result = grepResult.value;
       let fuzzyNotice: string | null = null;
 
-      // automatic fuzzy fallback allows to broad the queries and find different cases
-      if (result.items.length === 0 && !params.cursor && mode !== "regex") {
+      // automatic fuzzy fallback allows to broad the queries and find different cases.
+      // Skip it when the exact pass left a nextCursor: it did not exhaust the
+      // corpus (broad root / time budget hit), so a full fuzzy re-traversal would
+      // just repeat the expensive walk. Issue #746.
+      if (
+        result.items.length === 0 &&
+        !result.nextCursor &&
+        !params.cursor &&
+        mode !== "regex"
+      ) {
         // When the caller pinned a specific file (path has an extension), the
         // fuzzy fallback broadens across the whole picker — the file may just
         // be misnamed. For directory constraints (or no path), we keep the
@@ -738,6 +754,7 @@ export default function fffExtension(pi: ExtensionAPI) {
           beforeContext: 0,
           afterContext: 0,
           classifyDefinitions: true,
+          timeBudgetMs: GREP_TIME_BUDGET_MS,
         });
 
         if (fuzzy.ok && fuzzy.value.items.length > 0) {
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -36,11 +36,7 @@ const DEFAULT_FIND_LIMIT = 30;
 const GREP_MAX_LINE_LENGTH = 500;
 const MENTION_MAX_RESULTS = 20;
 
-// Native grep is synchronous, so an AbortSignal cannot interrupt a search
-// already in flight. Bound wall-clock time instead: over a broad root (e.g.
-// $HOME) an unbounded grep can scan for tens of minutes. 0 would mean
-// unlimited; the engine returns partial results plus a nextCursor on timeout.
-// Issue #746.
+// If we exceed 10 seconds for indexed grep - something is definetely off
 const GREP_TIME_BUDGET_MS = 10_000;
 
 type FffMode = "tools-and-ui" | "tools-only" | "override";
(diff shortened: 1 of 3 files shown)


packages/pi-fff/src/index.ts
```diff
@@ -36,6 +36,13 @@ const DEFAULT_FIND_LIMIT = 30;
 const GREP_MAX_LINE_LENGTH = 500;
 const MENTION_MAX_RESULTS = 20;
 
+// Native grep is synchronous, so an AbortSignal cannot interrupt a search
+// already in flight. Bound wall-clock time instead: over a broad root (e.g.
+// $HOME) an unbounded grep can scan for tens of minutes. 0 would mean
+// unlimited; the engine returns partial results plus a nextCursor on timeout.
+// Issue #746.
```
dmtrKovalenko:
```suggestion
// If we exceed 10 seconds for indexed grep - something is definetely off
```

packages/pi-fff/src/index.ts
```diff
@@ -713,15 +720,24 @@ export default function fffExtension(pi: ExtensionAPI) {
...
 
       let result = grepResult.value;
       let fuzzyNotice: string | null = null;
 
-      // automatic fuzzy fallback allows to broad the queries and find different cases
-      if (result.items.length === 0 && !params.cursor && mode !== "regex") {
+      // automatic fuzzy fallback allows to broad the queries and find different cases.
+      // Skip it when the exact pass left a nextCursor: it did not exhaust the
+      // corpus (broad root / time budget hit), so a full fuzzy re-traversal would
+      // just repeat the expensive walk. Issue #746.
```
dmtrKovalenko:
```suggestion
      // if we hit the timeout do not run the fuzzy fallback
      // cause it will only consumer more time 
```

dmtrKovalenko (approved):

packages/pi-fff/src/index.ts
```diff
@@ -36,6 +36,9 @@ const DEFAULT_FIND_LIMIT = 30;
 const GREP_MAX_LINE_LENGTH = 500;
 const MENTION_MAX_RESULTS = 20;
 
+// If we exceed 10 seconds for indexed grep - something is definetely off
```
dmtrKovalenko:
```suggestion
// If we exceed 10 seconds for indexed grep - something is definitely off
```

PR #753, opened 2026-08-08
Fix standalone constraints in multi_grep

### LINKED ISSUE #738: [Bug]: multi_grep drops standalone positive path constraints

Kh05ifr4nD:
Fixes #738.

Supersedes #739 with a configuration-level fix.

`multi_grep` receives patterns and constraints separately, but its adapters parsed the constraint field as an ordinary grep query. A lone path could become search text and disappear before the search. Exact-file misses could then discard the file constraint during prefiltering and search the wider index.

Parse the dedicated constraint field directly and keep multi-pattern prefiltering strict. Ordinary grep retains its existing query parsing and filename retry. The unsupported plain-grep fallback is removed.

Verified with the Rust workspace tests, Clippy, rustfmt, a release build, and MCP stdio cases covering directory, glob, exact and missing files, exclusions, and OR matching.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-pattern searches with file path and directory constraints.
  * Searches now correctly return no matches when constraints match no indexed files.
  * Prevented unmatched constraints from broadening search results.
  * Standardized empty multi-pattern results to display “0 matches.”
  * Applied consistent constraint handling across supported search interfaces.

* **Tests**
  * Added regression coverage for unmatched file path constraints and multiple standalone constraints.

### DIFF
--- a/crates/fff-core/src/grep/prefilter.rs
+++ b/crates/fff-core/src/grep/prefilter.rs
@@ -50,7 +50,7 @@ pub(super) fn prefilter_with_filepath_retry<'a>(
 /// Single pass prefilter that doesn't involve file reading
 /// allocates only amount of memory required for storing references of the FileItems have to be
 /// opened for grepping unaviodably, in the worst case allocates N * <word> memory if no prefilter needed
-fn prefilter_files<'a>(
+pub(super) fn prefilter_files<'a>(
     files: &'a [FileItem],
     constraints: &[Constraint<'_>],
     bigram_candidates: Option<&[u64]>,
--- a/crates/fff-core/src/grep/prefilter.rs
+++ b/crates/fff-core/src/grep/prefilter.rs
@@ -50,7 +50,7 @@ pub(super) fn prefilter_with_filepath_retry<'a>(
 /// Single pass prefilter that doesn't involve file reading
 /// allocates only amount of memory required for storing references of the FileItems have to be
 /// opened for grepping unaviodably, in the worst case allocates N * <word> memory if no prefilter needed
-pub(super) fn prefilter_files<'a>(
+pub(crate) fn prefilter_files<'a>(
     files: &'a [FileItem],
     constraints: &[Constraint<'_>],
     bigram_candidates: Option<&[u64]>,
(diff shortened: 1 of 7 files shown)


crates/fff-core/src/grep/prefilter.rs
```diff
@@ -50,7 +50,7 @@ pub(super) fn prefilter_with_filepath_retry<'a>(
 /// Single pass prefilter that doesn't involve file reading
 /// allocates only amount of memory required for storing references of the FileItems have to be
 /// opened for grepping unaviodably, in the worst case allocates N * <word> memory if no prefilter needed
-fn prefilter_files<'a>(
+pub(super) fn prefilter_files<'a>(
```
dmtrKovalenko:
```suggestion
pub(crate) fn prefilter_files<'a>(
```

dmtrKovalenko (approved):

PR #761, opened 2026-08-10
feat(fff.nvim): opt-in ordered fuzzy-part matching

yuhrao:
Adds an opt-in `file_picker.ordered_fuzzy_parts` setting (default off) so a multi-word fuzzy query like `foo bar` can be required to match as one in-order subsequence instead of each space-separated part matching independently anywhere in the candidate.

## Behavior

- **Off by default** — existing single/multi-word fuzzy search, grep (plain/regex/fuzzy), constraints (`*.rs`, `!test`, `/src/`, `type:`, `status:`), query history, and smart-case are all byte-for-byte unchanged.
- When enabled, valid (length-filtered) fuzzy parts are concatenated with **no separator** and routed through the existing single-needle match path (the same one already used for single-token queries) — no new matching algorithm. A space separator was tried first but empirically produced false negatives: a query with several short parts burns its whole typo budget on synthetic space characters that can never appear in a path, and the match is dropped entirely. Concatenating without a separator avoids that; see `test_ordered_fuzzy_parts_many_short_parts_still_match` and `test_ordered_fuzzy_parts_no_literal_space_required_in_candidate` in `score.rs`.
- Highlight recomputation (`fuzzy_match_byte_offsets_for_page`) and directory search share the same join helper (`join_ordered_parts`) and the same length filter as scoring, so highlights and pagination totals always agree with what actually matched.
- Fuzzy grep already joins its parts into a single needle via `FFFQuery::grep_text()` and is therefore already order-sensitive today, independent of this flag — no grep code changes were needed (locked down by `test_fuzzy_grep_multi_word_already_order_sensitive`). Plain-text and regex grep never look at per-part data at all, so they're structurally unaffected.
- Repeated whitespace between parts is a no-op — the parser already collapses it into discrete parts before any joining happens.

## API compatibility

- `FuzzySearchOptions` and `ScoringContext` are unchanged — no source break for `fff-c`, `fff-python`, `fff-mcp`, or the benches/bins.
- `FilePicker::fuzzy_search` / `fuzzy_search_directories` / `fuzzy_search_mixed` keep their exact existing signatures and always score unordered by default; new sibling `*_ordered` methods (explicit `bool` param) hold the actual logic, and are the only thing the Lua bindings call.
- Lua-facing `fuzzy_search_files` / `fuzzy_search_directories` / `fuzzy_search_mixed` gained a new **trailing** `Option<bool>` FFI argument, matching the existing pattern used for `live_grep`'s `smart_case`/`trim_whitespace` — old callers keep working unchanged.

## Tests

- Rust (`cargo test --workspace --no-default-features --features ripgrep --exclude fff-nvim`): all pass, plus new coverage in `score.rs` (ordered vs. unordered matching, repeated whitespace, constraints, smart-case bonus, pagination/highlight consistency, no-literal-space regression, many-short-parts separator-choice regression) and `grep_tests.rs` (order-sensitivity invariant).
- Lua (`PlenaryBustedDirectory tests/`): all 44 specs pass, including a new end-to-end spec (`tests/ordered_fuzzy_parts_spec.lua`) exercising the real FFI path against a built release binary.
- `cargo check -p fff-nvim --all-targets`: clean.

## Known limitation

I could not build or test with the `zlob` walker feature — it needs Zig, which wasn't available in my environment. All the Rust testing above used the default `ripgrep` walker feature instead. Flagging honestly in case CI (which does build `zlob`) catches something that path exposes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,7 @@ require('fff').setup({
   },
   file_picker = {
     fuzzy_query_highlighting = false, -- true to highlight fuzzy query matches in file picker results
+    ordered_fuzzy_parts = false, -- true to require space-separated query parts ("foo bar") to match in typed order
   },
   select = {
     -- Return winid to open the chosen file in, or nil to open in the original window
--- a/crates/fff-core/src/file_picker.rs
+++ b/crates/fff-core/src/file_picker.rs
@@ -1034,12 +1034,36 @@ impl FilePicker {
     /// The query should be parsed using [`crate::FFFQuery`] before calling
     /// this function. If a [`crate::QueryTracker`] is provided, the search will
     /// automatically look up the last selected file for this query and boost it
-    #[tracing::instrument(skip_all, name = "Fuzzy file search", fields(query = query.raw_query))]
     pub fn fuzzy_search<'q>(
         &self,
         query: &'q FFFQuery<'q>,
         query_tracker: Option<&QueryTracker>,
         options: FuzzySearchOptions<'q>,
+    ) -> SearchResult<'_> {
+        self.fuzzy_search_impl(query, query_tracker, options, false)
+    }
+
+    /// Same as [`FilePicker::fuzzy_search`], but opts into ordered fuzzy-part
+    /// matching: a multi-word query must match as one in-order subsequence
+    /// instead of each space-separated part matching independently anywhere
+    /// in the candidate.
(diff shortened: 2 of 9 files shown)


yuhrao:
Sorry, I left my agent working overnight for an experiment and it ended up doing more that I thought it would.

PR #793, opened 2026-08-17
fix(pi-fff): resolve startup config before tool registration

### LINKED ISSUE #792: [Bug]: pi-fff registers tools before startup mode is known
## What happens

`pi-fff` resolves its startup options and registers its tools while the extension is loading. At that point Pi has registered the extension flags, but has not populated their values yet. Saved `fff-mode` entries arrive later, in `session_start`, after the tool names have already been chosen.

On current `main` (`be2dd8d`), an isolated Pi session shows both sides of the problem:

```text
$ pi --fff-mode override
/fff-mode
Current mode: 'tools-and-ui' (flag: override)
```

The active extension tools are still `ffgrep` and `fffind`, rather than the expected `grep` and `find` overrides.

A session with a saved `override` selection has the opposite mismatch:

```text
Current mode: 'override' (flag: unset)
Active extension tools: ffgrep, fffind
```

This is the mismatch noted in the review of #790: https://github.com/dmtrKovalenko/fff/pull/790#discussion_r3792267379

The timing also affects the other public CLI flags. The database paths and scan options call the same `getConfigValue()` helper before Pi makes flag values available, so their CLI values are skipped during initial construction.

## Suggested fix

Keep flag registration where it is, but defer startup resolution and tool registration until `session_start`:

1. Resolve the five startup options after Pi has populated the flags.
2. Restore the saved session mode, preserving the current session-persistence behavior.
3. Create the finder factories and register the tools from that final state.
4. Explicitly activate the dynamically registered tool names.

For `/fff-mode`, switches between `tools-and-ui` and `tools-only` can remain immediate because their tool names are identical. A switch to or from `override` should save the selection but leave the reported active mode unchanged until `/reload`; that avoids another transient mismatch.

I have a focused patch ready on a fork. It keeps the tool definitions in place, queues only their final registration, and covers CLI precedence, saved-mode restoration, tool names, prompt guidelines, and the reload boundary. It passes the 74 `pi-fff` tests, type checking, formatting, and isolated Pi RPC checks. Happy to open the PR if this direction looks right.

XWIlluDelu:
## Summary

Pi applies extension flag values after the extension factory returns, while saved `fff-mode` entries become available in `session_start`. Resolve both there, then initialize the finders and register the tools with their final names. SDK sessions that do not bind `session_start` use the same one-time preparation before their first agent turn.

A `/fff-mode` change between prefixed and override tools is now saved but remains pending until `/reload`. Changes between `tools-and-ui` and `tools-only` still apply immediately. This keeps the reported mode, active tool names, and prompt guidelines in sync.

The tests now model Pi's flag lifecycle and cover startup precedence, invalid mode fallback, saved-mode restoration, dynamic registration, active-tool preservation, the SDK fallback, and the reload boundary.

## Validation

- `bun test packages/pi-fff/test` (76 tests)
- `bun run typecheck`
- Biome checks on the changed files
- Isolated `pi --mode rpc` checks for CLI override, invalid-mode fallback, and persisted override after reload
- An unbound `createAgentSession()` check confirming the tools are active before its first agent turn

Fixes #792.

## Summary by CodeRabbit

- **New Features**
  - Restores the selected mode when a session starts.
  - Applies compatible mode changes immediately and persists the selection.
  - Clearly indicates when `/reload` is required after switching to or from `override`.
  - Keeps available tools aligned with the restored or selected mode.
  - Reports initialization failures clearly and prevents unsupported access before setup completes.

- **Documentation**
  - Expanded `/fff-mode` guidance to cover mode precedence, session restoration, and reload requirements.

### DIFF
--- a/packages/pi-fff/README.md
+++ b/packages/pi-fff/README.md
@@ -116,7 +116,7 @@ Parameters:
 
 - `/fff-health` — show FFF status (indexed files, git info, frecency/history DB status)
 - `/fff-rescan` — trigger a file rescan
-- `/fff-mode <mode>` — switch mode (tool name change requires restart)
+- `/fff-mode <mode>` — switch mode (tool name changes require `/reload`)
 
 ## Modes
 
@@ -124,11 +124,13 @@ Parameters:
 - `tools-only`: additional tools only; keep pi's default `@` autocomplete
 - `override`: replaces pi's built-in `find`, `grep` and adds `multi_grep` + FFF-backed `@` autocomplete
 
-Mode precedence:
+Startup mode precedence:
 1. `--fff-mode <mode>` CLI flag
 2. `PI_FFF_MODE=<mode>` environment variable
 3. `mode` in the global config file
 4. default (`tools-and-ui`)
+
+When a session resumes, its most recent `/fff-mode` selection takes precedence over the startup resolution above. Switching to or from `override` takes effect after `/reload`, when the tools are registered again.
 
 ## Configuration
 
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -620,12 +620,12 @@ export default function fffExtension(pi: ExtensionAPI) {
   let toolsRegistered = false;
 
   function queueTool<TParams extends TSchema, TDetails = unknown, TState = any>(
-    name: keyof ToolNames,
+    resolveName: () => string,
     definition: PendingToolDefinition<TParams, TDetails, TState>,
   ): void {
     pendingTools.push(() => {
       const { promptGuidelines, ...tool } = definition;
-      const resolvedName = toolNames[name];
+      const resolvedName = resolveName();
       pi.registerTool({
         ...tool,
         name: resolvedName,
@@ -820,7 +820,7 @@ export default function fffExtension(pi: ExtensionAPI) {
     ),
   });
 
-  queueTool("grep", {
+  queueTool(() => toolNames.grep, {
     description: `Grep file contents. Smart-case, auto-detects regex vs literal, git-aware. Results are ranked by frecency (most-accessed files first); matches within a file stay in source order. Default limit ${DEFAULT_GREP_LIMIT}.`,
     promptSnippet: "Grep contents",
     promptGuidelines: (names) => [
@@ -1009,7 +1009,7 @@ export default function fffExtension(pi: ExtensionAPI) {
     ),
   });
 
-  queueTool("find", {
+  queueTool(() => toolNames.find, {
     description: `Fuzzy path search and glob search. Matches against the whole repo-relative path, not just the filename. Frecency-ranked, git-aware. Multi-word = narrower (AND). Default limit ${DEFAULT_FIND_LIMIT}.`,
     promptSnippet: "Find files by path or glob",
     promptGuidelines: (names) => [
@@ -1149,7 +1149,7 @@ export default function fffExtension(pi: ExtensionAPI) {
       cursor: Type.Optional(Type.String({ description: "Pagination cursor" })),
     });
 
-    queueTool("multiGrep", {
+    queueTool(() => toolNames.multiGrep, {
       description:
         "Search file contents for ANY of multiple literal patterns (OR, SIMD Aho-Corasick). Faster than regex alternation.",
       promptSnippet: "Multi-pattern OR content search",
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -6,7 +6,11 @@
  */
 
 import nodePath from "node:path";
-import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
+import type {
+  ExtensionAPI,
+  ExtensionContext,
+  ToolDefinition,
+} from "@earendil-works/pi-coding-agent";
 import {
   type AutocompleteItem,
   type AutocompleteProvider,
@@ -20,7 +24,7 @@ import type {
   MixedItem,
   SearchResult,
 } from "@ff-labs/fff-node";
-import { Type } from "@sinclair/typebox";
+import { Type, type TSchema } from "@sinclair/typebox";
 import { AuxFinderPool, routePathConstraint } from "./aux-finders";
 import { type FffMode, loadConfig, VALID_MODES } from "./config";
 import { FilePickerFactory } from "./file-picker";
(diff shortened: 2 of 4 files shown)


packages/pi-fff/README.md
```diff
@@ -116,20 +116,22 @@ Parameters:
...
 - `override`: replaces pi's built-in `find`, `grep` and adds `multi_grep` + FFF-backed `@` autocomplete
 
-Mode precedence:
+Startup mode precedence:
 1. `--fff-mode <mode>` CLI flag
 2. `PI_FFF_MODE=<mode>` environment variable
 3. `mode` in the global config file
 4. default (`tools-and-ui`)
 
+A mode saved with `/fff-mode` is restored when that session resumes. Switching to or from `override` takes effect after `/reload`, when the tools are registered again.
```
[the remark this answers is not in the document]
XWIlluDelu: Addressed in 9de4309. The README now states that the saved session selection takes precedence over startup resolution.

packages/pi-fff/src/index.ts
```diff
@@ -328,53 +331,57 @@ export default function fffExtension(pi: ExtensionAPI) {
...
+    currentMode = mode;
+    toolNames = resolveToolNames(mode);
+  }
 
-  function getMode(): FffMode {
-    return currentMode;
+  function resolveStartupConfig(): void {
+    setMode(
+      getConfigValue("fff-mode", "PI_FFF_MODE", config.mode, "tools-and-ui"),
+    );
```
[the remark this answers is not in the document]
XWIlluDelu: Addressed in 9de4309. `parseMode()` now rejects unsupported flag and environment values, allowing the shared resolver to continue to the next source. The fallback path is covered by a regression test.

packages/pi-fff/src/index.ts
```diff
@@ -731,16 +820,14 @@ export default function fffExtension(pi: ExtensionAPI) {
     ),
   });
 
-  pi.registerTool({
-    name: toolNames.grep,
-    label: toolNames.grep,
+  queueTool("grep", {
```
dmtrKovalenko: We should use toolNames.grep here

packages/pi-fff/src/index.ts
```diff
@@ -922,18 +1009,16 @@ export default function fffExtension(pi: ExtensionAPI) {
     ),
   });
 
-  pi.registerTool({
-    name: toolNames.find,
-    label: toolNames.find,
+  queueTool("find", {
```
dmtrKovalenko: ditto

packages/pi-fff/src/index.ts
```diff
@@ -1062,16 +1149,14 @@ export default function fffExtension(pi: ExtensionAPI) {
       cursor: Type.Optional(Type.String({ description: "Pagination cursor" })),
     });
 
-    pi.registerTool({
-      name: toolNames.multiGrep,
-      label: toolNames.multiGrep,
+    queueTool("multiGrep", {
```
dmtrKovalenko: ditto

packages/pi-fff/src/index.ts
XWIlluDelu: Updated in 039a8ea. The lazy getter references `toolNames.grep` directly while keeping name resolution deferred until startup state is available.

packages/pi-fff/src/index.ts
XWIlluDelu: Updated in 039a8ea.

packages/pi-fff/src/index.ts
XWIlluDelu: Updated in 039a8ea.

dmtrKovalenko (approved):

dmtrKovalenko:
@gustav-fff also you review and test out this PR

gustav-fff:
[triage-bot] DIRECTED: Reviewed and tested. Logic is sound, tests green. One real biome violation on the changed file — CI won't catch it (no biome step in `.github/workflows/`) but repo ships `biome.json` and PR body claims biome passed, so flagging.

**Tested** (branch checked out locally):
- `bun test packages/pi-fff/test` -> `76 pass, 0 fail`. Matches PR body.
- `bun run typecheck` -> only errors are `Cannot find module '@ff-labs/fff-node'` and the two implicit-`any` it cascades into (`index.ts:230,241`, both pre-existing, outside diff). Environmental: native workspace pkg not built locally. No type errors in changed code.
- Biome with pinned version (`bun.lock` -> `@biomejs/biome@2.5.8`): `biome check packages/pi-fff/src/index.ts` -> **2 errors**.

```
packages/pi-fff/src/index.ts:27:1 assist/source/organizeImports  FIXABLE
  × Sort the imported names.
packages/pi-fff/src/index.ts format
  × Formatter would have printed the following content
```

`biome check --write` produces:

```diff
-import { Type, type TSchema } from "@sinclair/typebox";
+import { type TSchema, Type } from "@sinclair/typebox";
@@ resolveStartupConfig() @@
-    setMode(
-      getConfigValue(
-        "fff-mode",
-        "PI_FFF_MODE",
-        config.mode,
-        "tools-and-ui",
-        parseMode,
-      ),
-    );
+    setMode(
+      getConfigValue("fff-mode", "PI_FFF_MODE", config.mode, "tools-and-ui", parseMode),
+    );
```

Fork PR, cannot push — apply the diff above (`cd packages && bunx biome check --write pi-fff/src/index.ts`).

**Logic**: startup ordering, `parseMode` fallthrough, saved-mode restore before registration, and the override `/reload` boundary all check out. Correctness hinges on `/reload` re-running the extension factory (resets `toolsRegistered`) — that is the load-bearing assumption; the immediate-vs-deferred split is correct given it.

Honk-Honk 🪿

dmtrKovalenko:
@gustav-fff make a separate PR with changing biome to oxc formatter and linter project wise and make sure that the they are correctly wired to the CI (only for packages/ folder)

gustav-fff:
[triage-bot] DIRECTED: Done — #796.

`biome` -> `oxfmt@0.63.0` + `oxlint@1.78.0`, scoped to `packages/`, and wired to CI (biome never was — that was the gap behind my earlier comment).

- `packages/.oxfmtrc.json` migrated with `oxfmt --migrate=biome`, so `printWidth: 90` / double quotes / trailing commas / semicolons are preserved. `ignorePatterns` holds the file set to `*.ts` like biome's `includes` did, so no README or `package.json` reflow noise.
- `packages/.oxlintrc.json`: `correctness` only. Wider categories produce ~45 findings across existing code (`no-await-in-loop`, `no-underscore-dangle`, `no-shadow`) — not this PR's job.
- Script names in `packages/package.json` unchanged, so `make format-ts` / `make lint-ts` / `make all` need no edit.
- Both `bun.lock` and `package-lock.json` regenerated — `external-tests.yml:151` runs `cd packages && npm ci`, stale lockfile would break it.
- `.github/workflows/oxc.yml` on `packages/**`, `-f github --deny-warnings`.

CI green on #796:

```
Check packages with oxfmt and oxlint	pass	12s
```

Verified failure path too: misformatted temp `.ts` -> `oxfmt --check` exit 1; duplicate object key -> `oxlint` exit 1 with `::error file=...` annotation. It is not a no-op job.

Two files reformatted, 13 lines: `packages/pi-fff/src/index.ts`, `packages/fff-node/src/index.ts`. Note oxc has no import sorter, so biome's `organizeImports` assist is dropped, not replaced — the `@sinclair/typebox` import order nit from my last comment just stops being a violation. If you want that enforced, it needs a separate tool.

Honk-Honk 🪿

PR #798, opened 2026-08-18
fix(fff-mcp): answer pre-initialize probes instead of exiting (#797)

### LINKED ISSUE #797: [Bug]: fff-mcp doesn't respond to Antigravity CLI's stateless MCP handshake ("server/discover")
### Which fff frontend?

MCP server (fff-mcp)

### has logs

_No response_

### Summary
fff-mcp fails to connect to Antigravity CLI. The client sends a `server/discover` 
request instead of the legacy `initialize` handshake, and fff-mcp closes the 
connection instead of responding, so the server never comes up.

### Error
✗ fff  error: Error: "Failed to start MCP server: expect initialized request, but received: 
Some(Request(JsonRpcRequest { jsonrpc: JsonRpcVersion2_0, id: Number(1), request: 
CustomRequest(CustomRequest { method: \"server/discover\", params: Some(Object {}), 
extensions: Extensions }) }))" : connection closed: calling "initialize": client is closing: EOF

### Root cause
Antigravity has adopted the newer stateless MCP spec (SEP-1442 / 2026-07-28 changelog), 
which drops the initialize/initialized handshake in favor of every request carrying its 
own protocol version, with server/discover as the first probe. fff-mcp still expects the 
old handshake and doesn't recognize server/discover, so it exits instead of returning a 
spec-compliant response.

Refs:
- https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442
- https://modelcontextprotocol.io/specification/2026-07-28/changelog

### Environment
- OS: Windows
- Client: Antigravity CLI 1.1.14
- fff-mcp: [run `fff-mcp --version` and paste output]
- Install method: install-mcp.ps1

### Expected
fff-mcp either supports the new stateless server/discover flow, or falls back 
gracefully to the legacy handshake when a client speaks the new spec (per SEP-1442, 
servers MAY support both).

### To reproduce
1. Install fff-mcp via install-mcp.ps1
2. Register it in Antigravity CLI's ~/.gemini/config/mcp_config.json
3. Run `agy`, check MCP Servers panel — fff shows the error above

gustav-fff:
Closes #797

## Root cause

Not Antigravity-specific. `rmcp` 1.7.0 (`service/server.rs:169-198`) accepts only `ping` before `initialize`; **any** other pre-init request returns `ExpectedInitializeRequest`, which `crates/fff-mcp/src/main.rs:343` turns into a hard startup error, exit code 1. A client that probes with `server/discover` therefore never gets a response and never gets a chance to fall back to the legacy handshake.

## Fix

New `crates/fff-mcp/src/handshake.rs`: `ProbeTolerantTransport` wraps the stdio transport and, while still uninitialized, answers unsupported requests with `-32601` and keeps reading instead of tearing down the connection. `initialize`, `ping`, notifications, and everything post-init pass through untouched — zero added work on the hot path (one `bool` check, and only until `initialize` arrives).

Scope note @dmtrKovalenko: this implements only the graceful-fallback half of the reporter's ask. Actually serving `server/discover` means implementing the SEP-1442 stateless flow, which is a new protocol mode and needs your scoping — `rmcp` 1.7.0 has no support for it (it decodes the method as `CustomRequest`).

## Steps to reproduce

On pre-fix `origin/main`:

```sh
cargo build -p fff-mcp
mkdir -p /tmp/fff-repro-797

# 1. discover probe alone
printf '{"jsonrpc":"2.0","id":1,"method":"server/discover","params":{}}\n' \
  | ./target/debug/fff-mcp --no-update-check /tmp/fff-repro-797; echo "EXIT=$?"

# 2. discover probe followed immediately by a valid initialize
{ printf '{"jsonrpc":"2.0","id":1,"method":"server/discover","params":{}}\n'
  printf '{"jsonrpc":"2.0","id":2,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"probe","version":"1"}}}\n'
  printf '{"jsonrpc":"2.0","method":"notifications/initialized"}\n'
  printf '{"jsonrpc":"2.0","id":3,"method":"tools/list","params":{}}\n'
  sleep 3; } | ./target/debug/fff-mcp --no-update-check /tmp/fff-repro-797; echo "EXIT=$?"
```

Expected: JSON-RPC error for the unknown method, server stays up, handshake completes, tools listed.

Actual on `main` — both cases exit 1 with nothing on stdout:

```
Error: "Failed to start MCP server: expect initialized request, but received: Some(Request(JsonRpcRequest { jsonrpc: JsonRpcVersion2_0, id: Number(1), request: CustomRequest(CustomRequest { method: \"server/discover\", params: Some(Object {}), extensions: Extensions }) }))"
```

## How verified

`cargo test -p fff-mcp` — 17 unit + 2 integration tests pass, including two new ones in `handshake.rs` (`pre_init_probe_is_rejected_and_initialize_still_arrives`, `post_init_requests_pass_through_untouched`).

`cargo clippy -p fff-mcp --all-targets` and `cargo fmt --check -p fff-mcp` — clean.

Repro case 2 above, post-fix:

```
id 1 ERROR -32601
id 2 result keys: ['capabilities', 'instructions', 'protocolVersion', 'serverInfo']
id 3 tools: ['find_files', 'grep', 'multi_grep']
EXIT=0
```

Regressions checked, all still `ok`: legacy `initialize` + `notifications/initialized` + `tools/call find_files`; pre-init `ping` followed by `initialize` (still answered by rmcp, not intercepted).

Not verified on Windows / real Antigravity CLI 1.1.14 — no such client here. Whether Antigravity actually retries with the legacy handshake after `-32601` is on @Divyesh172 to confirm.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup compatibility by tolerating early connection probes before initialization.
  * Unsupported pre-initialization requests now receive clear errors while the server continues waiting for initialization.
  * Initialization and ping requests are handled normally, with subsequent communication passed through unchanged.

### DIFF
--- a/crates/fff-mcp/src/handshake.rs
+++ b/crates/fff-mcp/src/handshake.rs
@@ -1,0 +1,184 @@
+use rmcp::model::{ClientRequest, ErrorCode, ErrorData, JsonRpcMessage};
+use rmcp::service::{RoleServer, RxJsonRpcMessage, TxJsonRpcMessage};
+use rmcp::transport::Transport;
+
+/// rmcp aborts startup on any pre-`initialize` request except `ping`, which kills the process
+/// before the client can retry. Answer such probes with `-32601` and keep waiting for the legacy
+/// handshake, so clients speaking the stateless spec (`server/discover`, SEP-1442) can fall back.
+/// @see https://github.com/dmtrKovalenko/fff/issues/797
+pub(crate) struct ProbeTolerantTransport<T> {
+    inner: T,
+    initialized: bool,
+}
+
+impl<T> ProbeTolerantTransport<T> {
+    pub(crate) fn new(inner: T) -> Self {
+        Self {
+            inner,
+            initialized: false,
+        }
+    }
+}
+
+impl<T> Transport<RoleServer> for ProbeTolerantTransport<T>
+where
+    T: Transport<RoleServer>,
+{
+    type Error = T::Error;
+
+    fn send(
+        &mut self,
+        item: TxJsonRpcMessage<RoleServer>,
+    ) -> impl Future<Output = Result<(), Self::Error>> + Send + 'static {
+        self.inner.send(item)
+    }
+
+    async fn receive(&mut self) -> Option<RxJsonRpcMessage<RoleServer>> {
+        loop {
+            let msg = self.inner.receive().await?;
+            if self.initialized {
+                return Some(msg);
+            }
+
+            let JsonRpcMessage::Request(request) = &msg else {
+                return Some(msg);
+            };
+
+            match &request.request {
+                ClientRequest::InitializeRequest(_) => {
+                    self.initialized = true;
+                    return Some(msg);
+                }
+                // rmcp answers pre-init pings itself
+                ClientRequest::PingRequest(_) => return Some(msg),
+                unsupported => {
+                    let error = unsupported_probe_error(unsupported.method());
+                    tracing::warn!(
+                        method = unsupported.method(),
+                        "rejecting pre-initialize request, awaiting initialize"
+                    );
+                    let id = request.id.clone();
+                    self.inner
+                        .send(JsonRpcMessage::error(error, Some(id)))
+                        .await
+                        .ok()?;
+                }
+            }
+        }
+    }
+
+    fn close(&mut self) -> impl Future<Output = Result<(), Self::Error>> + Send {
+        self.inner.close()
+    }
+}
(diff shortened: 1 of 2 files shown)


Divyesh172:
Confirmed on Antigravity CLI 1.1.14 / Windows 11: fff connects cleanly post-fix, find_files/grep return results.

PR #800, opened 2026-08-18
fix(core): recompute basename offset after lossy path decode (#799)

### LINKED ISSUE #799: [Bug]: Crash on invalid UTF-8 Unix filenames: lossy path decoding makes filename_offset slice inside U+FFFD
### Which fff frontend?

Node SDK (@ff-labs/fff-node)

### has logs

_No response_

### Description

FFF aborts while scanning a Unix directory containing a filename with invalid UTF-8 bytes.

The crash is fatal because it occurs on FFF’s Rayon background scan thread:

```text
thread 'fff-bg-0' panicked at crates/fff-core/src/file_picker.rs:2207:37:
end byte index 393 is not a char boundary; it is inside '�' (bytes 391..394 of string)
Rayon: detected unexpected panic; aborting
```
## Environment
- Linux x86_64 (glibc)
- @ff-labs/fff-node: 0.10.5
- @ff-labs/pi-fff: 0.10.5
- Pi coding agent: 0.84.2

## Reproducer

I attached fff-invalid-utf8-repro.tar.gz.

- SHA-256: fc0011038d27bbd63a3ca67a929d4f342b4895f922062fe86bd958712c67de08
- File: [fff-invalid-utf8-repro.tar.gz](https://github.com/user-attachments/files/31170039/fff-invalid-utf8-repro.tar.gz)
- Size: 10 KB
- Contents: directories plus one zero-byte file; no private file contents.

Extract it:

mkdir /tmp/fff-repro
tar -xf fff-invalid-utf8-repro.tar.gz -C /tmp/fff-repro

Then initialize @ff-labs/fff-node with /tmp/fff-repro as basePath:

const created = FileFinder.create({ basePath: "/tmp/fff-repro" });
if (!created.ok) throw new Error(created.error);

const finder = created.value;
await finder.waitForScan(10_000);
finder.destroy();

FFF aborts during the initial background scan.

gustav-fff:
Closes #799

## Root cause

`crates/fff-core/src/walk/zlob.rs:81-83` paired `entry.basename_offset_in_relative()` — measured on the **raw** path bytes — with `String::from_utf8_lossy(rel_bytes)`. Lossy decoding expands each invalid byte into a 3-byte U+FFFD, shifting the real separator forward, so the stored `filename_offset` can land inside a replacement char. `crates/fff-core/src/file_picker.rs:2205` then slices at it (`&rel[..filename_offset]`) and panics on the Rayon scan thread, which aborts the process. `FileItem::new_from_walk_bytes` (`file_picker.rs:498`) had the same defect. zlob-only: release/npm artifacts build `--features zlob`; the ripgrep backend derives the offset from the decoded string via `new_from_walk_parts` and is unaffected.

## Fix

New `decode_walk_rel_path()` helper decodes the bytes once and only recomputes the offset when the decode was actually lossy (`Cow::Owned`). Valid UTF-8 — every real-world path — takes the `Cow::Borrowed` arm and keeps the walker's offset, so the hot scan path is byte-for-byte the same work as before.

## Steps to reproduce

`\xff\xff` in a directory name cannot be created on APFS (`EILSEQ`), so the deterministic repro drives the exact walker output shape directly. On pre-fix `main`, add to `mod tests` in `crates/fff-core/src/file_picker.rs`:

```rust
#[test]
fn repro_799() {
    let raw: &[u8] = b"\xff\xffx/file.txt"; // separator at byte 3 -> zlob basename_offset = 4
    let (item, rel) = FileItem::new_from_walk_bytes(Path::new("file.txt"), raw, 4, None, 0, 0);
    let mut pairs = vec![(item, rel)];
    let walked = vec!["\u{FFFD}\u{FFFD}x/".to_string()];
    let mut builder = crate::simd_path::ChunkedPathStoreBuilder::new(1);
    let dirs = populates_dirs_files_chunked_storage(&mut pairs, &walked, &mut builder);
    let store = builder.finish();
    let arena = store.as_arena_ptr();
    let table: Vec<String> = dirs.iter().map(|d| d.relative_path(arena)).collect();
    assert_eq!(table, ["\u{FFFD}\u{FFFD}x/"]);
}
```

```sh
cargo test -p fff-search --no-default-features --features zlob --lib repro_799
```

Expected: `test result: ok`, dir table `["\u{FFFD}\u{FFFD}x/"]`, basename `file.txt`.

Actual on pre-fix `main`:

```text
thread 'file_picker::tests::repro_799' panicked at crates/fff-core/src/file_picker.rs:2205:37:
byte index 4 is not a char boundary; it is inside '\u{fffd}' (bytes 3..6) of `\u{fffd}\u{fffd}x/file.txt`
```

Full end-to-end variant on Linux/ext4 (matches the reporter's tarball):

```sh
python3 -c 'import os; d=b"/tmp/fff-repro/\xff\xffx"; os.makedirs(d, exist_ok=True); open(d+b"/file.txt","wb").close()'
# then FileFinder.create({ basePath: "/tmp/fff-repro" }) + waitForScan
# pre-fix: "Rayon: detected unexpected panic; aborting"
```

## How verified

```sh
cargo test --workspace --no-default-features --features zlob --exclude fff-nvim --exclude fff-python
# all 30 test binaries ok, 0 failed (includes the new dir_table_handles_invalid_utf8_dir_names)
cargo clippy --workspace --no-default-features --features zlob -- -D warnings   # clean
cargo check -p fff-search                                                       # ripgrep backend, clean
cargo fmt --all                                                                 # no diff
```

Regression test `dir_table_handles_invalid_utf8_dir_names` (`crates/fff-core/src/file_picker.rs`) fails with the char-boundary panic when the fix is reverted, passes with it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of files and directories with invalid UTF-8 characters in their names.
  * Preserved correct filename boundaries when displaying paths containing unusual characters.
  * Added regression coverage to ensure directory listings are built correctly for these paths.

### DIFF
--- a/crates/fff-core/src/file_picker.rs
+++ b/crates/fff-core/src/file_picker.rs
@@ -504,9 +504,7 @@ impl FileItem {
         modified: u64,
     ) -> (Self, String) {
         let is_binary = is_known_binary_extension(path);
-        // SAFETY-ish: paths on macOS/Linux are bytes; lossy conversion mirrors
-        // the existing `to_string_lossy()` behavior on non-UTF8 names.
-        let rel_str = String::from_utf8_lossy(relative_path).into_owned();
+        let (rel_str, basename_offset) = decode_walk_rel_path(relative_path, basename_offset);
         let item = Self::new_raw(basename_offset, size, modified, git_status, is_binary);
         (item, rel_str)
     }
@@ -2262,6 +2260,21 @@ fn push_dir_item(
     dirs.push(DirItem::new(dir_string, last_seg));
 }
 
+/// Decodes raw walker path bytes into the index's UTF-8 form, returning the
+/// basename offset valid for that string. Lossy decoding rewrites invalid
+/// bytes as 3-byte U+FFFD, so a raw-byte offset can land mid-char; recompute
+/// it in that (cold) case. Valid UTF-8 keeps the walker's offset for free.
+#[inline]
+pub(crate) fn decode_walk_rel_path(relative_path: &[u8], basename_offset: u16) -> (String, u16) {
+    match String::from_utf8_lossy(relative_path) {
+        std::borrow::Cow::Borrowed(valid) => (valid.to_owned(), basename_offset),
+        std::borrow::Cow::Owned(lossy) => {
+            let offset = lossy.rfind(std::path::is_separator).map_or(0, |i| i + 1) as u16;
+            (lossy, offset)
+        }
+    }
+}
+
 /// Fast extension-based binary detection. Avoids opening files during scan.
 /// Covers the vast majority of binary files in typical repositories.
 #[inline]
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
slop

PR #815, opened 2026-08-24
feat(nvim): add relative path display option

pompos02:
## Summary
Add a `file_picker.display_relative_path` configuration option, disabled by default, that renders `find_files` entries as a full picker-root-relative path.

This applies **only** to find_files. Grep results and grep suggestions retain their existing layout.
Long directories are shortened while keeping the filename visible. Git filename coloring and fuzzy match highlighting continue to use the correct offsets.

## Preview
**Before**
image

**After**
image

**Long Directories**
image

## Summary by CodeRabbit

* **New Features**
  * Added an option to display file-picker results using relative paths, such as `path/to/file`.
  * Relative-path display is controlled by `file_picker.display_relative_path` and is disabled by default.
  * File names, directories, and search highlights render correctly in both display formats.
  * Added configurable picker-input mappings, including an optional query-clearing action.

* **Documentation**
  * Documented file-picker open and close events.
  * Added sponsor information and a sponsorship contact link.

* **Tests**
  * Added coverage for relative-path rendering, narrow layouts, highlighting, and path shortening.


dmtrKovalenko (approved):
Thanks for a decent PR

dmtrKovalenko:
@gustav-fff i noticed that not all the configuration options are displayed inthe readme please ensure that they are in sync and make ordering more logical I think file_picker should be displayed right after or before layout to make it clear that it only applies to a file picker

dmtrKovalenko:
well well well I am not sure I like it now when I started thinking about configuration

1) it shouldn't be file picker only if user wants to display paths first it should be project wide config and should lie in layout
2) if we start configuring it per project we need a seperate module for rendering file names 
3) the dispay_relative_path is not 100% clear name cause they are displayed anyway, it's more like `show_path_first`

I reverted the commit so @gustav-fff can you please make a PR with the original commit from @pompos02 and apply all my suggested changes on top

gustav-fff:
[triage-bot] DIRECTED: opened #837.

@pompos02's original commit is the first commit, untouched. Second commit `e75e72a` applies your three points:

1. `file_picker.display_relative_path` -> `layout.show_path_first`. Dropped the `ctx.mode ~= 'grep' and ctx.suggestion_source ~= 'grep'` guard, so grep file group headers follow it too (they render through `file_renderer.render_line`).
2. New `lua/fff/picker_ui/file_name_renderer.lua` owns the name section — display text plus the byte offsets the git-status, directory and fuzzy extmarks need. `file_renderer.lua` loses 75 lines and no longer branches on the layout in four places.
3. Renamed to `show_path_first`.

Plus your earlier README ask: config block synced with `conf.lua` and `file_picker` moved right after `layout`. Missing keys added (`wrap_around`, `keymaps.cycle_forward_query`, `logging.enabled`, `file_picker.current_file_label`) and two wrong defaults fixed — `prompt` was documented as `'> '` but is `'🪿 '`, `layout.path_shorten_strategy` as `'middle_number'` but is `'middle'`.

No deprecation shim for `display_relative_path`, it never shipped in a release.

`make test-lua` 4/4 on `tests/file_renderer_spec.lua`, `make test-lua-snap` 42 cases 0 fails. The two reference screenshots from this PR are byte-identical after the rename, so the extraction is behaviour-preserving.

Honk-Honk 🪿

PR #818, opened 2026-08-25
feat(pi-fff): expose followSymlinks option

ohlulu:
Follow-up to #627 / #628: the option exists in the SDK, but the pi extension never passes it.

## Problem

FFF skips symlinks while walking, so any indexed tree that reaches its real files through links is invisible to pi-fff. Two layouts hit this:

- a git worktree whose `docs/` (or `.pi/`) is a symlink back into the main checkout, so the content has a single source of truth;
- the stowed dotfiles layout from #627, which is what motivated exposing the option in the first place.

In both cases `@`-mentions and the FFF-backed find/grep tools silently miss those files, while the same index rooted at the real directory finds them. Measured on a worktree here: `mixedSearch("repomap")` returns `[]` for the linked tree and `["docs/repomap.md"]` once `followSymlinks` is on.

#628 threaded `follow_symlinks` through `fff-c`, `fff-mcp`, `fff-python` and `@ff-labs/fff-node`, but `packages/pi-fff` was not part of it, so there is no way to reach the option from pi — not by flag, env, or `pi-fff.json`.

## Change

- `followSymlinks` joins the existing startup options and resolves through `getConfigValue()`, so precedence stays `--fff-follow-symlinks` > `FFF_FOLLOW_SYMLINKS` > `pi-fff.json` > default, exactly like the scanning options added in #790.
- Passed to both the cwd picker and the aux pickers. Only wiring the main finder would make absolute-path `fffind` / `ffgrep` disagree with workspace-relative queries about which files exist.
- `FffConfig`, the strict key check, the boolean validation, `pi-fff.schema.json` and the README config table / flag list all pick up the new key.

**Default is `true`** ([per review](https://github.com/dmtrKovalenko/fff/pull/818#issuecomment-5413518416)). A missing file has no failure mode an agent can see — it just gets an empty result and falls back to shell `find`/`grep`, which is the thing pi-fff exists to replace. `--fff-follow-symlinks=false` / `FFF_FOLLOW_SYMLINKS=0` / `"followSymlinks": false` are the opt-out. This diverges from `@ff-labs/fff-node`, the Rust core and fff.nvim, which stay `false`; pi-fff already diverges the same way on `enableHomeDirScanning`.

The first revision of this PR defaulted to `false` and repeated the cycle warning from `fff-c/src/ffi_types.rs` ("without external loop protection cyclic symlinks can wedge the watcher"). That warning is stale: the shipping zlob walker keeps a global `(dev, ino)` visited set (`zlob@v1.6.3 src/walker/worker.zig:42,123-171,360,643-647`, and `WalkFlags::FOLLOW_SYMLINKS` is documented as "cycles are detected and broken"), and the `ignore` fallback compares symlink targets against ancestor handles and returns `Error::Loop`, which `walk/ripgrep.rs:44-47` skips over. Cycles are the walker's problem, not the caller's, so the README no longer claims otherwise. The `ffi_types.rs` comment is left alone here — it belongs to a different package.

No `AGENTS.md`-style project config (`.pi/pi-fff.json`) is introduced here — the README already documents why the file is global only, and per-project scope would be a separate feature with its own trust story.

## Two behaviours worth a maintainer call

Neither is introduced by this PR, but defaulting to `true` widens both. Happy to bound either one here or in a follow-up.

- **Scope.** pi-fff enables `$HOME` scanning by default and #743 recorded a multi-hour, ~500 MB/s scan from a large `$HOME`. `ignore.rs` excludes hidden entries, `node_modules` and toolchain caches on non-git roots, but that is a filter, not a boundary — one directory symlink can now pull a large tree from outside `$HOME` or the workspace into the walk. `file_picker.rs:867-877` only guards the case where the picker base *is* `$HOME`.
- **Watcher coherence.** The initial scan follows symlinks, but the debouncer is fixed at `with_follow_symlinks(false)` (`background_watcher.rs:184-190`) while `index_new_directory` reads the picker's `follows_symlinks()` (`:717-735`). On macOS/Windows there is a single recursive watch on the base, so edits to a target outside the base may never reach the index after the first scan. Untested; Linux watches indexed directories individually and is probably fine.

## Test plan

```
$ cd packages/pi-fff && bun test test/
 81 pass, 0 fail, 204 expect() calls
```

Coverage:

- `config.test.ts` — `followSymlinks` round-trips through the full-option config, and `"true"` (string) is rejected as not a boolean.
- `extension.test.ts` — default `true` reaches `FileFinder.create`; `FFF_FOLLOW_SYMLINKS=0` turns it off; `"followSymlinks": false` in `pi-fff.json` turns it off; `--fff-follow-symlinks=false` beats `FFF_FOLLOW_SYMLINKS=1`, keeping the documented precedence.
- `aux-pool.test.ts` — an aux picker created for an out-of-workspace root inherits the setting.

```
$ cd packages && bunx oxfmt --check pi-fff/src pi-fff/test   # all matched files use the correct format
$ cd packages && bunx oxlint pi-fff/src pi-fff/test          # clean
```

`bun run typecheck` in `packages/pi-fff` reports the same six pre-existing errors on this branch as on an untouched `upstream/main` worktree (unresolved `@ff-labs/fff-node` types plus two implicit `any`s, all from the unbuilt native package) — no new ones.

Behaviour was also verified end-to-end against a real worktree with a symlinked `docs/`: with the extension patched to pass the option, a fresh `pi -p` session's `fffind repomap` returns `docs/repomap.md` as the top hit; without it, the tree is absent from the index.

## Summary by CodeRabbit

- **New Features**
  - Symlink traversal during indexing is now enabled by default.
  - Disable it with the `--fff-follow-symlinks` flag, environment variable, or configuration setting.
  - Auxiliary indexing and autocomplete honor the symlink traversal setting.
- **Documentation**
  - Updated configuration guidance to explain the new default and how to restrict indexing to the real directory tree.
- **Bug Fixes**
  - Improved validation for symlink traversal configuration values.

### DIFF
--- a/packages/pi-fff/README.md
+++ b/packages/pi-fff/README.md
@@ -145,7 +145,7 @@ For persistent global configuration, create `pi-fff.json` in pi's agent director
   "enableFsRootScanning": false,
   "enableHomeDirScanning": true,
   "warnOnHomeDirScan": true,
-  "followSymlinks": false
+  "followSymlinks": true
 }
 ```
 
@@ -160,7 +160,7 @@ All fields are optional:
 | `enableFsRootScanning` | boolean | `false` |
 | `enableHomeDirScanning` | boolean | `true` |
 | `warnOnHomeDirScan` | boolean | `true` |
-| `followSymlinks` | boolean | `false` |
+| `followSymlinks` | boolean | `true` |
 
 CLI flags take precedence over environment variables, which take precedence over this file. A missing file is ignored. Malformed JSON, unknown fields, and invalid values stop the extension from loading and report the file path and error. `/fff-mode` changes the current session; it does not edit this file.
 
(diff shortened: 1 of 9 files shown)


dmtrKovalenko:
oh wait this is not good we probably need to make it true by default

ohlulu:
Done — c2f4d81. The option stays; only pi-fff's fallback flips to `true`, with `--fff-follow-symlinks=false` / `FFF_FOLLOW_SYMLINKS=0` as the opt-out. The other frontends keep `false` since your comment was scoped here — say the word if you want those too.

Also dropped the README's cycle warning: I had copied it from `ffi_types.rs` and it is stale — zlob 1.6.3 keeps a global `(dev, ino)` visited set and the `ignore` fallback returns `Error::Loop`, so the walker breaks cycles itself.

Two pre-existing things the new default widens, both untouched here: `$HOME` scanning is on by default (#743) and one directory symlink can now pull an outside tree into that walk; and the watcher is fixed at `with_follow_symlinks(false)` (`background_watcher.rs:184-190`), so with macOS's single recursive watch, edits to a linked target outside the base may not reach the index. Happy to bound either one — your call.

PR #822, opened 2026-08-26
feat(pi-fff): add ui-only mode

pokedotdev:
## Why

   Some users run Pi with a custom harness where the FFF agent tools are unnecessary
   or may conflict with tools provided by the harness. File autocomplete with FFF
   is still useful in that setup.

   ## What changed

   This adds a new `ui-only` mode that:

   - Keeps `@` file autocomplete with FFF.
   - Does not register `fffind`, `ffgrep`, or `fff-multi-grep` as agent tools.
   - Can be selected through the existing flag, environment, config file, and
     `/fff-mode` options.
   - Requires `/reload` when switching to or from `ui-only`.
   - Leaves `tools-and-ui`, `tools-only`, and `override` unchanged.

   The mode is also included in the configuration schema, documentation, and tests.

   ## Testing

   - `cd packages/pi-fff && bun test test/`
   - `cd packages && bun format:check`
   - `make lint-ts`
   - `cd packages/pi-fff && PI_FFF_MULTIGREP=1 bun test test/extension.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Added a `ui-only` operating mode for FFF.
  * Enables FFF-backed `@`-file autocomplete without registering FFF agent tools.
  * Added `ui-only` to configuration options and `/fff-mode` commands.

* **Improvements**
  * Switching between `ui-only`, `override`, and tool-enabled modes now clearly indicates when `/reload` is required.
  * Updated documentation and configuration validation for the new mode.

* **Tests**
  * Added coverage for configuration support, autocomplete behavior, and reload requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,12 +133,13 @@ pi install npm:@ff-labs/pi-fff
 
 ### Modes
 
-Three operating modes, switchable at runtime with `/fff-mode`:
+Four operating modes, switchable at runtime with `/fff-mode`:
 
 | Mode                     | What it does                                                                      |
 | ------------------------ | --------------------------------------------------------------------------------- |
 | `tools-and-ui` (default) | Adds `ffgrep` and `fffind` tools, replaces `@`-mention autocomplete with FFF.     |
 | `tools-only`             | Only tool injection. Keeps pi's native editor autocomplete.                       |
+| `ui-only`                | Only `@`-file autocomplete with FFF; no FFF agent tools.                         |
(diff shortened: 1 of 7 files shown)


dmtrKovalenko:
I do not agree that it makes sense. 

1) fff has to be better by anything else custom implemented by the user
2) if user sure that their file search is better there very little value in adding this autocomplete so I’d recommend dropping fff at all

PR #827, opened 2026-08-28
fix(grep): enforce_time_budget option for zero-match searches (#826)

### LINKED ISSUE #826: [Bug]: grep timeBudgetMs is ignored for zero-match searches
### Which fff frontend?

Node SDK (@ff-labs/fff-node)

### has logs

_No response_

### Summary

`FileFinder.grep()` does not enforce `timeBudgetMs` when a plain-text search produces no matches.

In the reproduction below, a 5 ms search scans all 10,000 candidate files, returns `nextCursor: null`, and takes substantially longer than the requested budget.

Because the Node API is synchronous and exposes no cancellation mechanism, callers cannot independently interrupt the native operation. This makes `timeBudgetMs` the only available execution bound.

### Environment

- Linux x86-64
- Node.js 24
- Reproduced with `@ff-labs/fff-node` 0.9.4
- The relevant condition appears unchanged in the 0.10.5 source:
  https://github.com/dmtrKovalenko/fff/blob/v0.10.5/crates/fff-core/src/grep/grep.rs

### Reproduction

```sh
mkdir fff-budget-repro
cd fff-budget-repro
npm init -y
npm install @ff-labs/fff-node@0.10.5
```

Save the following as `repro.mjs`:

```js
import * as fs from "node:fs/promises";
import * as os from "node:os";
import * as path from "node:path";

function unwrap(result) {
  if (!result.ok) {
    throw new Error(`FFF operation failed: ${result.error}`);
  }
  return result.value;
}

const root = await fs.mkdtemp(
  path.join(os.tmpdir(), "fff-no-match-budget-"),
);

let finder;

try {
  const contents = `${"x".repeat(4 * 1024)}\n`;

  for (let batch = 0; batch < 100; batch += 1) {
    await Promise.all(
      Array.from({ length: 100 }, (_, index) =>
        fs.writeFile(
          path.join(root, `${batch}-${index}.txt`),
          contents,
        ),
      ),
    );
  }

  finder = unwrap(
    FileFinder.create({
      basePath: root,
      disableMmapCache: true,
      disableContentIndexing: false,
      aiMode: false,
      enableFsRootScanning: false,
      enableHomeDirScanning: false,
    }),
  );

  const ready = unwrap(await finder.waitForIndexReady(30_000));
  if (!ready) {
    throw new Error("Index did not become ready");
  }

  const startedAt = performance.now();

  const result = unwrap(
    finder.grep("needle-not-present", {
      mode: "plain",
      maxFileSize: 1024 * 1024,
      pageSize: 500,
      timeBudgetMs: 5,
    }),
  );

  const elapsedMs = performance.now() - startedAt;

  console.log({
    elapsedMs,
    totalFiles: result.totalFiles,
    filteredFileCount: result.filteredFileCount,
    totalFilesSearched: result.totalFilesSearched,
    nextCursor: result.nextCursor,
    cancellationAvailable: "cancel" in finder,
  });
} finally {
  finder?.destroy();
  await fs.rm(root, { recursive: true, force: true });
}
```

Run:

```sh
node repro.mjs
```

### Actual behavior

Example result from Linux x86-64:

```js
{
  elapsedMs: 59.25,
  totalFiles: 10000,
  filteredFileCount: 10000,
  totalFilesSearched: 10000,
  nextCursor: null,
  cancellationAvailable: false
}
```

The exact elapsed time is hardware-dependent, but the important behavior is that all candidate files are searched and no continuation cursor is returned after the time budget expires.

### Expected behavior

When `timeBudgetMs` expires:

- Grep should stop traversing candidate files even if it has found zero matches.
- `totalFilesSearched` should be less than `filteredFileCount`.
- `nextCursor` should allow the caller to resume from the next unsearched file, or the result should otherwise explicitly indicate that the search timed out.
- The synchronous native call should return within a small, documented amount of overshoot beyond the requested budget.

### Suspected root cause

The plain/regex grep budget check is conditional on existing matches:

```rust
if !need_abort
    && let Some(budget) = time_budget
    && all_matches.len() > 1
    && search_start.elapsed() > budget
{
    need_abort = true;
}
```

For a zero-match search, `all_matches.len()` remains zero, so the deadline condition is never applied.

Removing the match-count condition, while preserving cursor accounting for the partially consumed chunk, appears necessary.

### Relationship to #746 / #750

This is related to, but not resolved by:

- https://github.com/dmtrKovalenko/fff/issues/746
- https://github.com/dmtrKovalenko/fff/pull/750

PR #750 added `timeBudgetMs` to the `pi-fff` wrapper and deduplicated auxiliary index creation. It did not address the native zero-match budget bypass demonstrated here.

gustav-fff:
Closes #826

## Root cause

Two, both in the plain/regex path.

1. `crates/fff-core/src/grep/grep.rs` gated the deadline on `all_matches.len() > 1`, so a zero-match query never aborted.
2. Even with the gate lifted, `files_consumed` was unconditionally overwritten with `files_to_search.len()` whenever `result_files` was empty, forcing `next_file_offset = 0`. The cursor was lost. Same pattern in the fuzzy path at `crates/fff-core/src/grep/types.rs`.

## Fix

Per @dmtrKovalenko's review: (1) is behaviour existing callers depend on, so it is now opt-in rather than changed. (2) is a plain accounting bug and is fixed unconditionally.

- `GrepSearchOptions.enforce_time_budget`, default `false`. Off = historical rule, budget dormant until matches exist. On = hard bound, `next_file_offset` resumes at the first unsearched file.
- Surfaces: `enforceTimeBudget` (node/bun `GrepOptions`/`MultiGrepOptions`), `grep.enforce_time_budget` (nvim config, `content_search` opt), `enforce_time_budget=` (python kwarg).
- C ABI: `fff_live_grep` / `fff_multi_grep` keep their exact signatures and forward with `false`. The flag lives on new `fff_live_grep_ex` / `fff_multi_grep_ex` symbols. @dmtrKovalenko — appending a param to the existing symbols would have broken the frozen C ABI, hence the `_ex` pair rather than a 13th argument.
- Fuzzy grep already enforced the budget with zero matches; left alone, so the flag only affects plain/regex.
- An abort never skips file 0: `next_file_offset == 0` means "done" to every caller, so a page must consume at least one file or paging stalls.

Hot loop cost of the flag is zero — the gate is folded into the `Option<Duration>` before the rayon `par_iter`, not evaluated per file.

## Steps to reproduce

Both flag states are covered by one command on this branch — no patching needed. The `..._ignores_unenforced_...` case *is* pre-fix behaviour, asserted to still hold:

```sh
git checkout triage-bot/issue-826
cargo test -p fff-search --test grep_time_budget_zero_match --release -- --nocapture
```

10k files x 4KiB, `time_budget_ms: 5`, `page_limit: 500`, plain mode, query matches nothing.

- `enforce_time_budget: false` -> `total_files_searched == filtered_file_count == 10000`, `next_file_offset == 0`. Identical to `main`.
- `enforce_time_budget: true` -> `total_files_searched < 10000`, `next_file_offset > 0`.

For the pre-fix numbers on `main`, the reporter's `repro.mjs` in #826 runs unmodified. Rust-core equivalent measured on `main` at a51f0f4's parent:

```
elapsed_ms=111.93 total_files=10000 filtered_file_count=10000 total_files_searched=10000 next_file_offset=0
budget expired but all 10000 candidate files were searched
```

5ms requested, 111.93ms spent, whole candidate set scanned, no resume cursor.

Node, after `npm run build` in `packages/fff-node` — reporter's script with one line added:

```js
finder.grep("needle-not-present", {
  mode: "plain",
  maxFileSize: 1024 * 1024,
  pageSize: 500,
  timeBudgetMs: 5,
  enforceTimeBudget: true,
});
```

Expected: `totalFilesSearched < filteredFileCount`, `nextCursor` non-null, elapsed near 5ms. Drop `enforceTimeBudget` and the numbers are exactly what `main` prints today.

## How verified

```
$ cargo test -p fff-search --test grep_time_budget_zero_match --release
running 4 tests
test zero_match_fuzzy_search_stops_at_time_budget ... ok
test zero_match_search_ignores_unenforced_time_budget ... ok
test zero_match_search_stops_at_enforced_time_budget ... ok
test budget_resume_cursor_does_not_skip_files ... ok

test result: ok. 4 passed; 0 failed
```

`zero_match_search_ignores_unenforced_time_budget` is the regression guard for your concern: with the flag off, `total_files_searched == filtered_file_count` and `next_file_offset == 0`, exactly as before. `budget_resume_cursor_does_not_skip_files` pages a budget-limited search to exhaustion and asserts it still finds a needle planted in the last file, so the cursor drops nothing.

Also green: `cargo test -p fff-search --release --test grep_integration --lib grep` (18 tests), `cargo clippy --release -p fff-search -p fff-c -p fff-nvim --all-targets` (no new warnings), `cargo fmt`, `stylua`, `make header`, `make sync-js-api`.

Not verified: node/bun e2e and python bindings were not executed (no `node_modules` in this checkout) — those changes are signature plumbing only.

## Summary by CodeRabbit

- **New Features**
  - Added optional time-budget enforcement for grep and multi-grep searches across supported APIs.
  - Enforced budgets can stop searches before the first match and provide a cursor for resuming.
  - Interrupted and time-limited searches now resume from the precise stopping point.

- **Bug Fixes**
  - Prevented duplicate or skipped matches during pagination.
  - Improved “has more results” reporting and preserved progress for searches with no matches.

- **Documentation**
  - Documented the new time-budget enforcement setting and its behavior.


dmtrKovalenko:
https://github.com/dmtrKovalenko/fff/issues/826#issuecomment-5465015105 @gustav-fff work on this comment

gustav-fff:
[triage-bot] DIRECTED: already done — `8bf4031`, pushed before your mention landed.

Your premise is preserved as the default. `GrepSearchOptions.enforce_time_budget` defaults to `false`, and the budget gate is unchanged in that state:

```rust
// crates/fff-core/src/grep/grep.rs:601
let budget = time_budget.filter(|_| options.enforce_time_budget || all_matches.len() > 1);
```

Off = dormant until matches exist, exactly as before. Guarded by `zero_match_search_ignores_unenforced_time_budget`, which asserts `total_files_searched == filtered_file_count` and `next_file_offset == 0`.

Surfaces: `enforceTimeBudget` (node/bun), `grep.enforce_time_budget` (nvim config + `content_search` opt), `enforce_time_budget=` (python). C ABI untouched — `fff_live_grep`/`fff_multi_grep` keep their signatures and forward `false`; flag lives on new `fff_live_grep_ex`/`fff_multi_grep_ex`.

Zero hot-loop cost: the gate is folded into the `Option<Duration>` before the rayon `par_iter`, not evaluated per file.

The second half of the PR is a separate bug and stays unconditional: `files_consumed` was overwritten with `files_to_search.len()` whenever `result_files` was empty, forcing `next_file_offset = 0` and losing the cursor even on a match-bearing abort.

`be0c61c` on top: style-only, moved the test helpers to end of file per CLAUDE.md.

```
$ cargo test -p fff-search --test grep_time_budget_zero_match --release
test result: ok. 4 passed; 0 failed
```

Also: CodeRabbit flagged `crates/fff-mcp/src/server.rs` losing the cursor on an empty page. Unreachable — MCP passes `time_budget_ms: 0`, `enforce_time_budget: false`, `abort_signal: None` (server.rs:62-68), so it never aborts, and `next_file_offset` is returned in the struct at server.rs:402/648 regardless. `"0 matches."` is only the display string. No change made.

Honk-Honk 🪿

PR #831, opened 2026-08-28
fix: keep grep path constraint when the pinned file exists (#830)

### LINKED ISSUE #830: ffgrep: path constraint targeting an unindexed file silently widens to a repo-wide fuzzy search
### Summary

`ffgrep` with a `path` targeting a **file outside the index** (e.g. anything under `node_modules/`, which is gitignore/exclusion-respected) returns zero exact matches, then silently re-runs the search **repo-wide with the `path` constraint dropped**, presenting unrelated files as if they were related to the pinned path.

### Environment

- OS: Windows 11
- pi: 0.84.3
- @ff-labs/pi-fff: 0.10.5 (tools-and-ui mode)
- CWD: Node project with `node_modules/` (react installed); `node_modules/react/package.json` exists on disk and actually contains `jsx-runtime` in its `exports`.

### Reproduction

All commands run from the repo root:

1. `ffgrep { pattern: "jsx-runtime", path: "node_modules/react/package.json" }`
   -> `[0 exact matches. Maybe you meant this?]` then matches from `package-lock.json` and `tests/` — never from the pinned file (or anything under `node_modules`).

2. `ffgrep { pattern: "version", path: "node_modules/react/package.json" }`
   -> first result is `e2e/file-link-repro.spec.ts:50: version: 3,` — completely unrelated to `react`.

3. `ffgrep { pattern: "sideEffects", path: "node_modules/react/package.json" }`
   -> returns `src/renderer/**` files whose top hit is a `useEffect` line — fuzzy noise with zero relationship to either the query or the pinned path. (Because results are frecency-ranked, the `App.tsx` family shows up first for generic words.)

Control cases for contrast:

- `path: "node_modules/react/"` (directory constraint, no trailing extension) + `jsx-runtime` -> clean `No matches found` (no leak — directory constraints keep the constrained query).
- `path: "this-dir-does-not-exist-xyz/"` -> clean `No matches found`.
- `path: "package-lock.json"` (an indexed file) -> works correctly.

### Root cause

In `src/index.ts` (the grep tool's `execute`):

- Workspace-relative paths never reach an aux finder: `routePathConstraint` (`aux-finders.ts`) returns `null` for paths without a `../` prefix, so the query runs against the **workspace finder**, whose index excludes `node_modules`. Exact grep => 0 items.
- On 0 items, `pathTargetsFile` is computed from the trailing extension, and when true the fuzzy fallback uses `fuzzyQuery = pattern` **without the path constraint** (the "file may just be misnamed" heuristic):

```ts
const lastSeg = params.path?.split(/[\/]/).pop() ?? "";
const pathTargetsFile = /\.[a-zA-Z][a-zA-Z0-9]{0,9}$/.test(lastSeg);
const fuzzyQuery = pathTargetsFile ? pattern : query;
const fuzzy = picker.grep(fuzzyQuery, { mode: "fuzzy", ... });
```

The heuristic never checks whether the pinned file is actually reachable in the picker/index. For any unindexed file path — and `node_modules/...` is always unindexed — every grep is **guaranteed** to hit this fallback and return repo-wide noise.

### Why it's dangerous

- An agent pinning `path: "node_modules/<pkg>/package.json"` to check a dependency's contents receives matches from arbitrary source files, plausibly presented as if they concerned that file.
- Short, common patterns (`version`, `sideEffects`, ...) amplify the problem: the fuzzy results are barely related to the pattern at all.

### Expected behavior

1. **Don't broaden when the target isn't in the index.** If `pathTargetsFile` and the pinned path is not part of the picker's index (gitignored/excluded), return a clear message such as `path not indexed (gitignored or excluded): no matches` — or, better, read/match that single file on disk so the query actually answers the question.
2. **At minimum, state that the constraint was dropped.** When the fallback does broaden repo-wide, the output must say: `path constraint ignored (target not indexed); results are repo-wide`.
3. The fuzzy fallback for common/generic patterns degrades into pure noise (`sideEffects` -> `useEffect` files). Consider suppressing or clearly labeling fuzzy results.

### Notes

- Directory constraints (no trailing extension) keep the constrained query and are safe — verified `No matches found`, no leak.
- Aux finders currently can't rescue this case because `routePathConstraint` refuses workspace-relative paths even when they point outside the indexed set (e.g. `node_modules/react/...`).

gustav-fff:
Closes #830

## Root cause

`packages/pi-fff/src/index.ts:930-932` decided to drop the `path` constraint from the fuzzy fallback purely from the trailing extension of `params.path`, never checking that the pinned file was reachable in the picker. Since `routePathConstraint` (`packages/pi-fff/src/aux-finders.ts:165`) keeps workspace-relative paths on the workspace finder, and that index excludes `node_modules/**`, every grep pinned there returned 0 exact matches and then re-ran repo-wide with the constraint gone.

## Fix

Widening is now limited to the case the heuristic was written for: a `path` that resolves to nothing on disk (misnamed guess), and the notice says the constraint was dropped. A pinned file that exists keeps its constraint; if it is also absent from the index, the notice states that instead of returning unrelated files.

## Steps to reproduce

```sh
git checkout main && make prepare-bun
mkdir -p /tmp/fff830 && cd /tmp/fff830 && git init -q
printf 'node_modules/\n' > .gitignore
mkdir -p node_modules/react src/renderer
printf '{"name":"react","exports":{"./jsx-runtime":"x"}}' > node_modules/react/package.json
printf 'export const App = () => { useEffect(() => {}, []) }\n' > src/renderer/App.tsx
```

Then, with the checked-out extension loaded (`pi -e packages/pi-fff/src/index.ts -p --no-session ...`), call:

```
ffgrep { pattern: "jsx-runtime", path: "node_modules/react/package.json" }
```

expected — the pinned path is honoured:

```
[node_modules/react/package.json is not indexed (gitignored or excluded)]
No matches found
```

actual on `main` — constraint silently dropped, unrelated file returned:

```
[0 exact matches. Maybe you meant this?]
src/renderer/App.tsx
 12: useEffect(() => {
```

The same three cases are covered deterministically by the new tests in `packages/pi-fff/test/extension.test.ts` (`pi-fff grep fuzzy fallback`); they fail on `main` with exactly the output above and pass here.

## How verified

```
83 pass, 0 fail (was 80 pass before the 3 new + 1 control test)
```

Against the real native picker on the fixture above (temporary harness, not committed): `glob("package.json")` returns `package.json`, `glob("node_modules/react/package.json")` returns `[]`, and `ffgrep` with the pinned path yields the `not indexed` message for `jsx-runtime` and `sideEffects` instead of `src/renderer/App.tsx`.

`bunx oxfmt` and `bunx oxlint packages/pi-fff` clean. The extra `glob` pass runs only after an exact grep already returned zero rows, so no hot query gains work.

@dmtrKovalenko one unrelated thing surfaced while verifying: with `path: "node_modules/react/package.json"` and `pattern: "version"`, the *exact* grep still matched the workspace-root `package.json`. That is `normalizePathConstraint` emitting a FilePath constraint that matches on basename, not the fallback, and it lives in the query parser — left alone here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fuzzy search handling for file path constraints.
  * Existing files that are not indexed now retain their path constraint and receive a clear notification.
  * Missing file paths now trigger a repository-wide fuzzy search with an explanation that the constraint was dropped.
  * Directory-based constraints continue to work correctly.

* **Tests**
  * Added coverage for indexed, unindexed, missing, and directory path scenarios.

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -5,6 +5,7 @@
  * @-mention autocomplete suggestions to the interactive editor.
  */
 
+import fs from "node:fs";
 import nodePath from "node:path";
 import type {
   ExtensionAPI,
@@ -28,7 +29,7 @@ import { Type, type TSchema } from "@sinclair/typebox";
 import { AuxFinderPool, routePathConstraint } from "./aux-finders";
 import { type FffMode, loadConfig, VALID_MODES } from "./config";
 import { FilePickerFactory } from "./file-picker";
-import { isHomeDir, resolveDbPaths } from "./paths";
+import { HOME_DIR, isHomeDir, resolveDbPaths } from "./paths";
 import { buildQuery } from "./query";
 
 export { SCAN_TIMEOUT_MS } from "./sdk";
@@ -929,7 +930,14 @@ export default function fffExtension(pi: ExtensionAPI) {
         // excluded / out-of-scope directories.
         const lastSeg = params.path?.split(/[\\/]/).pop() ?? "";
         const pathTargetsFile = /\.[a-zA-Z][a-zA-Z0-9]{0,9}$/.test(lastSeg);
-        const fuzzyQuery = pathTargetsFile ? pattern : query;
+        // A pinned file that exists on disk is not a misnamed guess, so widening
+        // past it would answer a different question with repo-wide noise (#830).
+        const pinnedFile =
+          pathTargetsFile && params.path
+            ? resolvePinnedFile(params.path, activeCwd)
+            : null;
+        const dropConstraint = pathTargetsFile && !pinnedFile;
+        const fuzzyQuery = dropConstraint ? pattern : query;
         const fuzzy = picker.grep(fuzzyQuery, {
           mode: "fuzzy",
           smartCase,
@@ -943,8 +951,15 @@ export default function fffExtension(pi: ExtensionAPI) {
         });
 
         if (fuzzy.ok && fuzzy.value.items.length > 0) {
-          fuzzyNotice = `0 exact matches. Maybe you meant this?`;
+          fuzzyNotice = dropConstraint
+            ? `0 exact matches, path constraint dropped (no such file: ${params.path}); results are repo-wide`
+            : `0 exact matches. Maybe you meant this?`;
           result = fuzzy.value;
+        } else if (
+          pinnedFile &&
+          !isIndexedFile(picker, aux ? aux.root : activeCwd, pinnedFile)
+        ) {
+          fuzzyNotice = `${params.path} is not indexed (gitignored or excluded)`;
         }
       }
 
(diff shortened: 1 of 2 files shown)


mjisanshak6-glitch:
Hello bro

PR #836, opened 2026-08-29
fix(pi-fff): avoid blocking startup on initial scan

leezenn:
## Summary

- return the native finder immediately after successful creation
- let the initial filesystem scan continue in the background
- add regression coverage preventing `waitForScan()` from returning to the shared factory path

## Why

Pi emits `session_shutdown` followed by `session_start` during `/reload`. The shared picker factory currently waits up to 15 seconds for the initial scan, so every reload can block on a fresh traversal.

The finder is usable while scanning; searches may temporarily return partial results. This trade-off avoids blocking Pi startup and reload while preserving scan progress reporting and lifecycle cleanup.

## Verification

- `cd packages/pi-fff && bun test test/` — 81 tests passed
- focused live Pi reload A/B: stock 3.24 s; patched 30–114 ms
- immediate search while `isScanning` was true returned a valid result

## Summary by CodeRabbit

* **Improvements**
  * File pickers are now available immediately while scanning continues in the background.
  * Reduced startup waiting time when creating a file picker.

* **Tests**
  * Added coverage confirming picker creation does not wait for the initial scan to finish.

### DIFF
--- a/packages/pi-fff/src/file-picker.ts
+++ b/packages/pi-fff/src/file-picker.ts
@@ -1,5 +1,5 @@
 import type { FileFinderApi, InitOptions, Result } from "@ff-labs/fff-node";
-import { type FileFinderStatic, loadSdk, SCAN_TIMEOUT_MS } from "./sdk";
+import { type FileFinderStatic, loadSdk } from "./sdk";
 
 export interface PickerOptions {
   basePath: string;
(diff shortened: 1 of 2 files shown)


leezenn:
excuse me and my сlanker for this mess
сорян 😄

PR #845, opened 2026-08-31
fix(fff-mcp): omit git:clean from find_files output

### LINKED ISSUE #844: [Suggestion]: omit git:clean annotations from fff-mcp output
### Which fff frontend(s)?

MCP server (fff-mcp)

### Summary

`fff-mcp` currently appends `git:clean` to clean files returned by `find_files`, for example:

```text
src/a.ts git:clean
src/b.ts git:clean
src/c.ts git:modified
```

I think clean files should have no Git annotation, while dirty states such as `modified`, `untracked`, `deleted`, and staged states should continue to be shown.

### Why this matters

`clean` is the overwhelmingly common/default state in a repository. Repeating it for nearly every result adds visual noise and consumes unnecessary LLM context without adding much information.

The absence of a Git annotation can already imply the default state, similar to how `git status --short` only surfaces files that differ from the default state.

For an MCP tool used by coding agents, the higher-value information is the exceptional/actionable state:

```text
src/a.ts
src/b.ts
src/c.ts git:modified
src/new.ts git:untracked
```

This preserves the useful Git signal while making the output more compact and easier for models to scan.

### Existing inconsistency in this repository

`pi-fff` already follows this behavior in `fffFileAnnotation()`: it explicitly suppresses `clean`, `unknown`, and empty Git states, and only surfaces actionable Git state.

By contrast, `fff-mcp` uses `format_git_status_opt()` inside `file_suffix()`. Since `format_git_status_opt(None)` returns `Some("clean")`, `file_suffix()` ends up materializing the default state as `git:clean` for every clean file.

There is also already a useful separation in the core implementation: during the initial scan, `git_status: None` effectively means clean, so clean is represented internally as the absence of an exceptional state. The MCP presentation layer could preserve that idea instead of turning it back into a visible annotation.

### Expected behavior

Keep full Git status information in the core/query layer so constraints such as `git:modified` continue to work, but suppress `clean` in the MCP presentation layer.

Conceptually, `fff-mcp` could treat the formatted Git status as absent when it is `clean`:

```rust
let git_status = format_git_status_opt(git_status)
    .filter(|status| *status != "clean");
```

Then only non-default Git states would be appended to file results.

I think this would be better as the default MCP behavior rather than a new configuration option, because `clean` is the default state and hiding it loses no actionable information. A verbose/debug mode could still expose the complete status if needed.

### Proposed solution

_No response_

gustav-fff:
Closes #844

## Root cause

`format_git_status_opt(None)` returns `Some("clean")` (`crates/fff-core/src/git.rs:142-144`), so `file_suffix()` (`crates/fff-mcp/src/output.rs:23`) materialized the default git state as ` git:clean` on every `find_files` result line (`crates/fff-mcp/src/server.rs:537`). `pi-fff` already suppresses it (`packages/pi-fff/src/index.ts:160`).

## Fix

One filter in the MCP presentation layer only. Core keeps the full status, so `git:modified` / `git:untracked` query constraints are unaffected. Dirty and staged states still render.

```rust
format_git_status_opt(git_status).filter(|status| *status != "clean"),
```

## Steps to reproduce

```sh
cargo build --release -p fff-mcp

# fixture: 8 files, one modified, one staged, one untracked
rm -rf /tmp/fff-844 && mkdir -p /tmp/fff-844/src && cd /tmp/fff-844
git init -q . && git config user.email b@b && git config user.name b
for n in a b c d e f g h; do printf 'export const %s = 1\n' $n > src/svc_$n.ts; done
git add -A && git commit -qm init
printf 'export const a = 2\n' > src/svc_a.ts
printf 'export const b = 3\n' > src/svc_b.ts && git add src/svc_b.ts
printf 'export const n = 1\n' > src/svc_new.ts

# then call find_files(query="svc") over stdio against this dir
```

Expected (this PR) vs actual on pre-fix `main`:

```text
  expected                         actual (main)
  src/svc_a.ts git:modified        src/svc_a.ts git:modified
  src/svc_b.ts git:staged_modified src/svc_b.ts git:staged_modified
  src/svc_new.ts git:untracked     src/svc_new.ts git:untracked
  src/svc_d.ts                     src/svc_d.ts git:clean
  src/svc_e.ts                     src/svc_e.ts git:clean
  ...                              ...
```

## How verified

- `cargo test -p fff-mcp` green, plus 2 new unit tests in `output.rs` pinning that `clean`/`CURRENT`/empty/`None` produce no tag while dirty + staged states and frecency words survive.
- `cargo clippy -p fff-mcp` clean.
- Live stdio probe of the built binary on the fixture above — output matches the table.
- `scripts/benchmark-claude.sh` A/B, baseline binary vs this one, fff arm, 11 concepts each: task completion **10/11 in both arms**, 0 errors in both. Concept 2 fails in both (its target path no longer exists in the corpus — stale ground truth, unrelated). Cost/turn deltas are pure noise (turn-count variance swamps the change): $1.8214 baseline vs $1.7151 patched.
- Context saved where it applies: on a 20-result `find_files` page over `chromium`, ` git:clean` was 200 of ~1600 chars, i.e. **-12.4%** of payload. The eval suite itself only calls `find_files` 3-6 times across 11 concepts (grep dominates: 21 calls), so suite-level byte savings are small — the win shows up in find_files-heavy sessions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clean files no longer display an unnecessary Git status annotation.
  * Git status indicators continue to appear for modified, untracked, and staged files.

### DIFF
--- a/crates/fff-mcp/src/output.rs
+++ b/crates/fff-mcp/src/output.rs
@@ -21,9 +21,11 @@ fn frecency_word(score: i32) -> Option<&'static str> {
 }
 
 pub fn file_suffix(git_status: Option<git2::Status>, frecency_score: i32) -> String {
+    // `clean` is the default state of nearly every file: annotating it burns
+    // context without adding signal. Only exceptional git state is surfaced.
     match (
         frecency_word(frecency_score),
-        format_git_status_opt(git_status),
+        format_git_status_opt(git_status).filter(|status| *status != "clean"),
     ) {
         (Some(f), Some(g)) => format!(" - {f} git:{g}"),
         (Some(f), None) => format!(" - {f}"),
@@ -544,6 +546,27 @@ fn collect_file_preview<'a>(
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn suffix_omits_clean_git_status() {
+        assert_eq!(file_suffix(None, 0), "");
+        assert_eq!(file_suffix(Some(git2::Status::CURRENT), 0), "");
+        assert_eq!(file_suffix(Some(git2::Status::empty()), 0), "");
+        assert_eq!(file_suffix(None, 120), " - hot");
+    }
+
+    #[test]
+    fn suffix_keeps_dirty_git_status() {
+        assert_eq!(
+            file_suffix(Some(git2::Status::WT_MODIFIED), 0),
+            " git:modified"
+        );
+        assert_eq!(file_suffix(Some(git2::Status::WT_NEW), 0), " git:untracked");
+        assert_eq!(
+            file_suffix(Some(git2::Status::INDEX_MODIFIED), 60),
+            " - warm git:staged_modified"
+        );
+    }
 
     #[test]
     fn trunc_strips_trailing_whitespace() {


dmtrKovalenko:
@gustav-fff whats wrong with the CI?

gustav-fff:
[triage-bot] DIRECTED: Not this PR. Exactly one job failed — `Python bindings (macos-latest)` — on a flaky fff-python grep pagination test.

```
tests/test_finder.py::test_grep_cursor_paginates_by_file
>           assert second.total_matched >= 1
E           assert 0 >= 1
E            +  where 0 = GrepResult(items=0, total_matched=0, next_file_offset=0).total_matched
1 failed, 27 passed in 12.26s
```

Rerun of the identical SHA `3e3596a` is green: https://github.com/dmtrKovalenko/fff/actions/runs/33409845165. Everything else was green on the first pass, including python on ubuntu/windows. Diff touches only `crates/fff-mcp/src/output.rs`; `fff-python` does not depend on `fff-mcp`.

Mechanism, hypothesis — 15/15 local passes on macOS, so it needs runner load: the cursor is a plain numeric offset into the prefiltered candidate list, which is re-sorted on every call (`crates/fff-core/src/grep/prefilter.rs:171`) then sliced with `split_off` (`:178`), and files without a bigram entry are appended to the tail (`:134-145`). `wait_for_scan_blocking()` waits for the file scan, not for content indexing, so page 1's offset can be computed against a different ordering than page 2 sees and page 2 skips the only other file matching `def`. Pagination is ordering-dependent by construction — separate issue if you want it fixed.

Honk-Honk 🪿

PR #856, opened 2026-09-07
fix(pi-fff): drop stale mode tool names on reload (#855)

### LINKED ISSUE #855: pi-fff: switching modes leaves stale tool names active after `/reload`
## Summary

`registerPendingTools` only ever adds names to the active tool set, never removes old ones. After switching from `override` back to a FFF-named mode and running `/reload`, the restored active list still contains the previous mode's tool names, so the model sees duplicate search tools.

Per the README, `override` **replaces** the built-in `find` / `grep`, and switching modes takes effect after `/reload` when the tools are registered again. The stale names contradict both statements.

## Reproduction

1. Start Pi with `pi-fff` in `override` mode (`grep` / `find` active).
2. Run `/fff-mode tools-and-ui`, then `/reload`.
3. Check the active tools (e.g. via `/tool` or the model's tool list).

## Actual

Stale search tools stay active alongside the final mode's names: `grep`, `find`, `ffgrep`, `fffind` (functionally duplicated in pairs).

## Expected

The active set matches the final mode: only `ffgrep` / `fffind` (plus `fff-multi-grep` when enabled).

## Verification (two-session mock on #854's branch)

Fresh extension instance per reload, one shared persisted active list:

| direction | active set after reload | status |
|---|---|---|
| default -> override | `read, grep, find` | fixed by #854's prune |
| override -> default | `read, grep, find, ffgrep, fffind` | still open (this issue) |

## Notes

- The default-to-override direction was fixed as a side effect of #854's prune: pre-registered FFF names are instance-tracked, so they are removed from the active set.
- Fresh sessions are unaffected in either direction.
- A fix needs to distinguish the extension's `grep` / `find` registrations from Pi's built-ins sharing those names (e.g. via `getAllTools` source info), so it is kept out of #854.

gustav-fff:
Closes #855

## Root cause

`registerPendingTools` was purely additive (`packages/pi-fff/src/index.ts:659-665`), while pi hands the pre-reload active list straight back to the rebuilt tool registry (`dist/core/agent-session.js:2063`: `activeToolNames: this.getActiveToolNames()`). A `/reload` after a mode switch therefore keeps the previous mode's names active, and both directions are affected on `main`.

Relevant detail: pi's default active set is `["read", "bash", "edit", "write"]` (`dist/core/sdk.js:132`) — builtin `grep`/`find` are registered but inactive. A leftover `grep`/`find` in a FFF-named mode is genuinely the extension's stale activation.

## Fix

`registerPendingTools` now takes the names the final mode did not register and drops them from the active set. FFF names (`ffgrep`/`fffind`/`fff-multi-grep`) are pruned unconditionally — no one else registers them. Override names (`grep`/`find`/`multi_grep`) are pruned only when this session actually ran in override mode, determined from the startup-resolved mode plus every persisted `/fff-mode` entry, so a builtin `grep`/`find` the user enabled via `defaultTools` survives.

## Steps to reproduce

```sh
cd packages && bun install --frozen-lockfile
git checkout triage-bot/issue-855 -- pi-fff/test/extension.test.ts
cd pi-fff && bun test test/extension.test.ts
```

On pre-fix `main`:

```
(fail) pi-fff mode switch across /reload > drops override tool names when switching back to a FFF-named mode
(fail) pi-fff mode switch across /reload > drops FFF tool names when switching to override
 26 pass
 2 fail
```

Expected active tools after override -> `tools-and-ui` + `/reload`: `["read","bash","edit","write","ffgrep","fffind"]`.
Actual on `main`: `["read","bash","edit","write","grep","find","ffgrep","fffind"]` — four search tools, duplicated in pairs.

Manual equivalent (TUI):

```
pi -e packages/pi-fff/src/index.ts --fff-mode=override
/fff-mode tools-and-ui
/reload
/tool          # main: grep, find, ffgrep, fffind active. fixed: ffgrep, fffind
```

## How verified

```sh
cd packages/pi-fff && bun test test/     # 85 pass, 0 fail
cd packages && bun run format:check      # all matched files use the correct format
cd packages && bun run lint              # oxlint clean
cd packages/pi-fff && bun run typecheck  # no new errors
```

Three tests added modelling pi's real cross-reload behaviour (active list and session entries survive, extension instance does not): both switch directions, plus a guard that a user-enabled builtin `grep`/`find` is not pruned when override was never used. `typecheck` still reports the two pre-existing `TS7006` in `src/index.ts` and the unbuilt `@ff-labs/fff-node` resolution errors — identical on `main`.

@dmtrKovalenko one judgment call worth your eyes: telling the extension's `grep`/`find` from pi's builtins is impossible by name, so the prune leans on mode history as the evidence. If you would rather have hard provenance, the alternative is persisting the activated names in a session entry.

Overlaps #854, which touches the same function — the prune there becomes redundant with this one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `/reload` mode switches so tools from a previous mode no longer remain active.
  - Preserved user-enabled built-in tools, including `grep` and `find`, when reloading sessions or opting out of home-directory scanning.
  - Improved restoration of persisted session modes to keep the active tool list accurate.
  - Added clearer warnings when scanning is unavailable because a directory has opted out.

### DIFF
--- a/packages/pi-fff/src/index.ts
+++ b/packages/pi-fff/src/index.ts
@@ -78,6 +78,10 @@ function resolveToolNames(mode: FffMode): ToolNames {
   return mode === "override" ? OVERRIDE_TOOL_NAMES : FFF_TOOL_NAMES;
 }
 
+function toolNameList(names: ToolNames): string[] {
+  return [names.grep, names.find, names.multiGrep];
+}
+
 // ---------------------------------------------------------------------------
 // Cursor store — simple bounded Map for pagination cursors
 // ---------------------------------------------------------------------------
@@ -656,11 +660,19 @@ export default function fffExtension(pi: ExtensionAPI) {
     });
   }
 
-  function registerPendingTools(): void {
+  // Pi carries the active tool list across /reload, so names activated under a
+  // previously used mode stay active unless we drop them here (#855).
+  function registerPendingTools(staleNames: readonly string[]): void {
     if (toolsRegistered) return;
 
-    const registeredNames = pendingTools.map((register) => register());
-    pi.setActiveTools([...new Set([...pi.getActiveTools(), ...registeredNames])]);
+    const registeredNames = new Set(pendingTools.map((register) => register()));
+    const stale = new Set(staleNames.filter((name) => !registeredNames.has(name)));
+    pi.setActiveTools([
+      ...new Set([
+        ...pi.getActiveTools().filter((name) => !stale.has(name)),
+        ...registeredNames,
+      ]),
+    ]);
     toolsRegistered = true;
   }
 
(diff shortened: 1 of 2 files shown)


dmtrKovalenko:
@gustav-fff rebase this PR

gustav-fff:
[triage-bot] DIRECTED: rebased onto `main` (`e542d8a`), force-pushed as `f56030f`. `mergeStateStatus: CLEAN`.

One conflict, in `prepareSession` (`packages/pi-fff/src/index.ts`), against #858. Both sides kept, but they fight:

- #858 keeps FFF names in `override` when the cwd is opted out of indexing, so pi's builtin `grep`/`find` stay usable.
- this PR prunes override names from the active set when the session ever ran in `override`.

Composed naively that prune deactivates exactly the builtins #858 just preserved:

```
override + $HOME + FFF_ENABLE_HOME_SCAN=0
  before: read bash edit write grep find
  after:  read bash edit write ffgrep fffind     # grep/find gone, nothing searches
```

Resolution: prune override names only when FFF actually takes those names.

```ts
const keepBuiltins = currentMode === "override" && scanOptOutReason(activeCwd) !== null;
if (keepBuiltins) toolNames = FFF_TOOL_NAMES;
if (usedOverride && !keepBuiltins) staleNames.push(...toolNameList(OVERRIDE_TOOL_NAMES));
```

Added one test pinning it (`test/extension.test.ts:499`); drop the `!keepBuiltins` and it fails with the diff above.

`bun test test/` 93 pass 0 fail, `format:check` clean, `lint` clean, `typecheck` unchanged from `main`.

#854 is still not on `main`, so the overlap note in the PR body stands.

Honk-Honk 🪿

PR #863, opened 2026-09-09
fix(core): stop labelling unclassified git statuses as clean

kevin9327:
## The defect

A merge-conflicted file is reported as `clean`.

`format_git_status_opt` ends with:

```rust
} else if status.contains(Status::CURRENT) || status.is_empty() {
    Some("clean")
} else {
    None
}
```

`GIT_STATUS_CURRENT` is `0` in libgit2, so `git2::Status::CURRENT` is the empty
bitflag and `status.contains(Status::CURRENT)` is vacuously true for **every**
status. Every state the chain does not name explicitly therefore falls out as
`clean`, and the `else { None }` arm — plus `format_git_status`'s
`unwrap_or("unknown")` — is unreachable.

The states that hit it: `CONFLICTED`, `WT_TYPECHANGE`, `INDEX_TYPECHANGE`,
`INDEX_RENAMED`.

`CONFLICTED` genuinely reaches this function: `fuzz_git_watcher_stress.rs`
already asserts "picker must surface CONFLICTED after merge". So after a
conflicted merge the picker shows those files as clean, and since #845 the MCP
output drops the annotation entirely rather than flagging anything.

The Lua layer is already set up for the fallback — `GIT_LABELS` in
`lua/fff/file_picker/file_info.lua` carries `unknown = '?'`, a label the Rust
side could never produce.

## The fix

Test `is_empty()` and let anything else fall through to the existing `unknown`
fallback, as the code was written to do.

`RUSTUP_TOOLCHAIN` was pinned to `1.98.0-x86_64-pc-windows-msvc` because
`rust-toolchain.toml`'s `stable` channel could not update on this machine.

Before, `cargo test -p fff-search --lib -- git::`:

```
test git::tests::unclassified_status_is_not_reported_as_clean ... FAILED

---- git::tests::unclassified_status_is_not_reported_as_clean stdout ----

thread 'git::tests::unclassified_status_is_not_reported_as_clean' (41116) panicked at crates\fff-core\src\git.rs:198:9:
assertion `left == right` failed
  left: Some("clean")
 right: None

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 158 filtered out
```

After:

```
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 158 filtered out
```

The same test pins every label that must not move — `None`, `Status::empty()`
and `Status::CURRENT` all still read `clean`, and `untracked` / `modified` /
`staged_modified` / `ignored` are unchanged. Those assertions pass both before
and after, so only the previously-misfiled states move.

- `cargo test -p fff-search --lib` — 160 passed, 0 failed.
- `cargo test -p fff-mcp --bins` — 22 passed, 0 failed (covers
  `suffix_omits_clean_git_status` and `suffix_keeps_dirty_git_status`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Git status reporting now correctly identifies conflicted and type-changed files instead of incorrectly labeling them as clean.
  - Existing recognized Git status labels remain unchanged.

### DIFF
--- a/crates/fff-core/src/git.rs
+++ b/crates/fff-core/src/git.rs
@@ -159,7 +159,9 @@ pub fn format_git_status_opt(status: Option<Status>) -> Option<&'static str> {
                 Some("staged_deleted")
             } else if status.contains(Status::IGNORED) {
                 Some("ignored")
-            } else if status.contains(Status::CURRENT) || status.is_empty() {
+            // `Status::CURRENT` is 0, so `contains` is true for every status:
+            // testing it here labelled conflicted/typechanged files "clean".
+            } else if status.is_empty() {
                 Some("clean")
             } else {
                 None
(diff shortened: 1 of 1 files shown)


dmtrKovalenko (approved):
yes lol this makes more sense

PR #866, opened 2026-09-09
Fix pi-fff worktree path constraints

pb-sameerreddy:
## Summary

Fixes Pi FFF query scoping and adds configurable default excludes:

- preserves recursive directory globs like `.predibase-buildenv/**` instead of collapsing them to `.predibase-buildenv/`
- adds a `defaultExcludes` config option for path constraints that should be excluded from every `find` / `grep` query
- skips those default excludes when the query path explicitly targets one of them

## Why

The current normalization collapses `dir/**` to `dir/`. In FFF query syntax that directory-prefix constraint can match the same directory segment anywhere in the indexed tree, so a query scoped to `.predibase-buildenv/**` can also match `.worktrees/foo/.predibase-buildenv/**`.

Configurable default excludes let users avoid stale or generated subtrees, such as local git worktree roots, without baking product-specific paths into the package. Explicit searches inside those paths still work.

## Example config

```json
{
  "defaultExcludes": [".worktrees/", ".claude/worktrees/"]
}
```

## Validation

- `cd packages && bun run format:check`
- `cd packages/pi-fff && bun test`

Note: `cd packages/pi-fff && bun run typecheck` currently fails before this change on missing `@ff-labs/fff-node` type resolution plus existing implicit-any errors in `src/index.ts`.

## Summary by CodeRabbit

- **New Features**
  - Added configurable default exclusions for find and grep queries.
  - Queries automatically exclude configured paths unless searching within an excluded location.

- **Bug Fixes**
  - Recursive directory patterns now retain their original glob format for more precise matching.
  - Exclusion patterns are normalized consistently, improving query results for paths such as `test/**`.

### DIFF
--- a/packages/pi-fff/pi-fff.schema.json
+++ b/packages/pi-fff/pi-fff.schema.json
@@ -46,6 +46,12 @@
       "type": "boolean",
       "default": true,
       "description": "Indexes through directory symlinks, e.g. a git worktree or stow layout whose files live behind links. Set to false to keep the walk inside the real tree."
+    },
+    "defaultExcludes": {
+      "type": "array",
+      "items": { "type": "string", "minLength": 1 },
+      "default": [],
+      "description": "Path constraints to exclude from every find and grep query unless the query path explicitly targets one of them."
     }
   }
 }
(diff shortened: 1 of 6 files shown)


dmtrKovalenko (commented):
There is an option in the file walking that allow to exlclude some files from the indexer itself. But I am not sure your motivation is correct

If you are trying to hide some results from llm here are some questions:

1. why? 
2. have you tried `.ignore` (not .gitignore) file as a project based configuration to hide it
3. why do you think it is a good idea to have a global configuration 

we can make this happen but I feel like there might be some misunderstanding

dmtrKovalenko:
closing becaus of inactivity after my questions

